PR #35, opened 2015-07-02
Index

ebuchman:
- Refactoring so anything referencing cobra and all flags are dealt with in the `commands` package
- Add support for numbered containers, using `--num` on `data`, `services`, `chains`
- `stop` and `rm` take the `-x` flag for removing associated data container
- Various cleanups


chains/manage.go
```diff
@@ -456,19 +337,14 @@ func RenameChainRaw(oldName, newName string) error {
...
 	//   variable will mean it pulls. But we want the opposite default
 	//   behaviour for chains as we do for services in this regard
 	//   so we flip the variable.
-	var skipPull bool
-	if pull {
-		skipPull = false
-	} else {
-		skipPull = true
-	}
+	skipPull := !pull
```
compleatang: neat. didn't know I could do that.

util/docker_client.go
```diff
@@ -53,3 +54,45 @@ func DockerConnect(verbose bool) {
 		}
 	}
 }
+
+func NameAndNumber(name string, number int) string {
```
compleatang: the Prisoner of War function ;)

PR #77, opened 2015-07-07
wasn't compiling, added run bool to setupChain

zramsay:


chains/manage.go
```diff
@@ -295,7 +295,7 @@ func CatChainRaw(chainName string) error {
 
 // the main function for setting up a chain container
 // handles both "new" and "fetch" - most of the differentiating logic is in the container
-func setupChain(chainID, chainName, cmd, dir, genesis, config string, containerNumber int, publishAllPorts bool) (err error) {
+func setupChain(chainID, chainName, cmd, dir, genesis, config string, containerNumber int, publishAllPorts bool, run bool) (err error) {
 	// XXX: if chainName is unique, we can safely assume (and we probably should) that containerNumber = 1
```
ebuchman: when you have two variables in a row in a go function of the same type, you only need to mention the type one. See how the the first 6 args are strings? The last two are bool ....

ebuchman:
sorry i dealt with this

PR #81, opened 2015-07-09
the do struct

compleatang:
The Do Struct.

All of the Raw functions have now been reconfigured to have the following signature:

```go
func XXXXRaw(do *definitions.Do) error {
...
}
```

The Do struct may need some massaging, but it encapsulates everything that all of the Raws need to do their job. This greatly simplifies testing and working on things.


ebuchman:
#hero

Issue #74, opened 2015-07-07
`files cache`


compleatang:
make IPFS `pin` things. by hash

compleatang:
let's get you up to speed with ipfs :)

PR #96, opened 2015-07-12
Files cache #74, in progress

zramsay:
Basic functionality is there. Needs: handle the Headers -> better error msgs.
TBD: the IPFSBaseUrl() ... had to change it to go through the api. `pin` via POST through gateway not possible.


files/handle.go
```diff
@@ -77,6 +76,39 @@ func PutFiles(do *definitions.Do) error {
...
+func PinFiles(do *definitions.Do) error {
+	var hash string
+	var err error
+
+	ipfsService, err := loaders.LoadServiceDefinition("ipfs", 1)
+	if err != nil {
+		return err
+	}
+
+	if services.IsServiceRunning(ipfsService.Service, ipfsService.Operations) {
```
compleatang: We have tons of this code now. How do you feel about abstracting this sequence into a

`services.EnsureRunning(do)` 

which we would call rather than adding these long if statements everywhere?

util/writers.go
```diff
@@ -31,6 +32,54 @@ func SendToIPFS(fileName string, w io.Writer) (string, error) {
...
+	body, err := PinAPICall(url, fileHash, w)
+	if err != nil {
+		return "", err
+	}
+	fmt.Printf("response: %v", body)
+	return body, nil
+
+}
+
+func PinAPICall(url, fileHash string, w io.Writer) (string, error) {
```
compleatang: let's abstract this one step (naming wise) since we're already passing a url, let's call this function `PostAPICall` and make it for HTTP-Post routes and just formulate the URL in the `PinToIPFS` function. 

that will give us more flexibility going forward for other api calls we need to make.

files/handle.go
zramsay: makes sense. I'm on it.

util/writers.go
zramsay: was just about to start that ... finally got the pinning working properly.

util/writers.go
```diff
@@ -31,6 +33,69 @@ func SendToIPFS(fileName string, w io.Writer) (string, error) {
...
+	var errs struct {
+		Message string
+		Code    int
+	}
+	var p struct {
+		Pinned []string
+	}
+	var msg string
+
+	if response.StatusCode >= http.StatusBadRequest {
```
zramsay: bit of a hack from: https://github.com/ipfs/go-ipfs/blob/master/commands/http/client.go#L221
needs better handling, esp. with PostAPICall abstraction ... or maybe deal with responses within the func making the call ... and PostAPI can handle the errs

zramsay:
clsoing for now while working on #98

Issue #67, opened 2015-07-05
build container number automagician to find a new free containerNumber


compleatang:
will require some massaging of the container naming regex

compleatang:
the foundation for names and containerNumber should be all set. 

two things need to happen:

* need to add a bit of autoMagic to the containerNumbers (as in if no containerNumber is given, find the first free container number)
* need to hook up from the Flag parsing step to the Load/Mock step

... then we should be good.

@zramsay I think this will be a good one for you. see also my comments on #69 (first bullet point in chains i think).

compleatang:
There are a couple of ways to do this. We probably want to build this automagician to run during the Ops overwrite functionality when the operations struct which is assembled by the CLI level is used to override any default operations populated during the loading process:

https://github.com/eris-ltd/eris-cli/blob/develop/services/services_test.go#L106

The function should check to see if ContainerNumber is 0||nil and if it is, it should set it to 1 for now. We may want to add in a bool that will tell the function whether it should be unique (this is a new service container) or whether it should use the highest number found (for when we need to find an existing container, like for stopping and such)

zramsay:
TODO
https://github.com/eris-ltd/eris-cli/blob/develop/commands/services.go#L325
 (still not sure if mod to BuildServicesGroup is legit).

zramsay:
^nvm, already taken care of.

PR #100, opened 2015-07-14
Automagic #67

zramsay:
The magician works. Brute forcing my way around tests...still need to figure out what's going on there.
To make `--num` work for services I had to mod BuildServicesGroup, maybe that breaks something?


chains/operate.go
```diff
@@ -39,6 +39,7 @@ func InstallChain(do *definitions.Do) error {
 
 func StartChain(do *definitions.Do) error {
 	logger.Infoln("Ensuring Key Server is Started.")
+	//should it take a flag? keys server may be running another cNum
 	keysService, err := loaders.LoadServiceDefinition("keys", 1)
```
compleatang: let's use `EnsureRunning` here.

services/operate.go
```diff
@@ -2,37 +2,36 @@ package services
...
-func StartService(do *definitions.Do) error {
+func StartService(do *definitions.Do) (err error) {
 	var services []*definitions.ServiceDefinition
-	if do.Operations.ContainerNumber == 0 { // TODO: automagic, see #67
-		do.Operations.ContainerNumber = 1
-	}
 
+	cNum := do.Operations.ContainerNumber
+	if err != nil {
+		return err
```
compleatang: FYI, with named returns in go: `(err error)` (in function signature) then you would just add `return` here....

services/operate.go
```diff
@@ -2,37 +2,36 @@ package services
...
-		s, e := BuildServicesGroup(srv)
+		s, e := BuildServicesGroup(srv, cNum)
 		if e != nil {
 			return e
 		}
 		services = append(services, s...)
 	}
-
 	for _, s := range services {
+		//XXX does AutoMagic elim need for this?
```
compleatang: No, cause this OverWriteOps call makes us zero out (or harmonize) the ops struct prior to running, which is the effect that we want....

PR #101, opened 2015-07-14
rm services/chains exec, closes #94

### LINKED ISSUE #94: remove services exec and chains exec as executable commands

zramsay:


ebuchman:
I'm wondering if we should keep services exec. Only because I realized that `tinydns` isn't using a data container in the standard mindy. But I should probably `gofmt` myself and give it one if I want to be able to exec.

Thing is - the tinydns container is used in `--volumes-from` for the mindy container, so it's like a data container, but its running a service too (and starts with `eris_service`) - I wonder if we can just stick a flag on `eris data exec` to allow execing on service containers too, or if I should really `gofmt` myself.

zramsay:
lmk what you guys decide and I'll update the PR

compleatang:
I think we really only want `exec` to be ran against data containers. The ui could prove a point of confusion (as in, do I `services start` or `services exec` for noobs). In addition, I think @ebuchman 's instincts here were right the first time. Am going to merge this in as is. If we really do need to have services exec let's revisit it "on the next turn" (as lawyers often say).

PR #102, opened 2015-07-17
Services restart, closes #95

### LINKED ISSUE #95: `services restart`

zramsay:


zramsay:
either I don't yet understand tests (likely) or circle is broken (also likely). diff of those two commits have nothing to do with chains but it's TestRename/KillChain that fails

PR #118, opened 2015-08-03
Chains graduate test, closes #92

### LINKED ISSUE #92: chains graduate test

zramsay:
works. but now `TestNewService` fails ... not sure how much refactoring I should do (& what else it'll break) to make that test pass.


definitions/service_definition.go
```diff
@@ -6,25 +6,34 @@ type ServiceDefinition struct {
...
+	Service     *Service     `json:"service" yaml:"service" toml:"service"`
+	ServiceDeps *ServiceDeps `json:"services" json:"services,omitempty", yaml:"services,omitempty" toml:"services,omitempty"`
+	Maintainer  *Maintainer  `json:"maintainer,omitempty" yaml:"maintainer,omitempty" toml:"maintainer,omitempty"`
+	Location    *Location    `json:"location,omitempty" yaml:"location,omitempty" toml:"location,omitempty"`
+	Machine     *Machine     `json:"machine,omitempty" yaml:"machine,omitempty" toml:"machine,omitempty"`
+	Srvs        []*Service
+	Operations  *Operation
+}
+
+type ServiceDeps struct {
```
compleatang: This is a bit clunky and causing panics in the marshalling process. We can clean that up, but wondering your thinking for breaking these out into their own struct instead of just marshalling the array as before?

zramsay: The array was throwing an error writing to .toml, which expects a struct.

Issue #112, opened 2015-07-26
`--pull-images` flag for `eris init`


compleatang:
* eris/keys
* eris/ipfs
* eris/erisdb
* eris/eth

if `--dev`

* erisindustries/evm_compilers
* eris/compilers
* erisindustries/node
* erisindustries/python
* erisindustries/gulp
* erisindustries/embark_base
* erisindustries/sunit_base
* erisindustries/pyepm_base

ebuchman:
lol who's using that python image?!

zramsay:
so E pointed out that this feature may be overkill, since the default behaviour of `docker run` includes `pull`. Tests also take 2-3x as long.

compleatang:
fair. we can add a --no-pull flag then to squelch that behaviour.

zramsay:
`--no-pull` set to true when testing on circle (otherwise it fails)

zramsay:
Somewhere down the line this feature got axed. I'm re-opening to suggest a `--pull-images[=false]` flag that, when true, pulls:
- `eris/base`
- `eris/keys`
- `eris/data`
- `eris/erisdb`
- `eris/epm`
because these are all needed for the pipeline from key/chain making to deploying contracts & once `init` has been run, waiting on images for each of three commands breaks the flow

compleatang:
bumping to 0.12

zramsay:
closing in favour of task for #309

Issue #91, opened 2015-07-11
eris ls


compleatang:
list all services, chains, data containers.

(eventually workers too)

PR #119, opened 2015-08-04
autopop mntnr info; `eris known, ls`; pull images on init; closes #68, #91, #112

### LINKED ISSUE #68: auto populate maintainer information on `chains new` and `services new`
use git config to populate.

zramsay:


util/init.go
```diff
@@ -28,6 +28,38 @@ func Initialize(toPull, verbose bool) error {
...
+	fmt.Println("Pulling base images...")
+
+	if !dev {
+		argsAll = argsDef
+	} else {
+		fmt.Println("...and development images")
+		argsAll = append(argsDef, argsDev...)
+	}
+
+	fmt.Println("This could take awhile, now is a good time to feed your marmot")
```
compleatang: :+1:

util/init.go
```diff
@@ -28,6 +28,38 @@ func Initialize(toPull, verbose bool) error {
...
+	for _, img := range argsAll {
+		pImg := []string{"pull", img}
+		err := exec.Command("docker", pImg...).Run()
```
compleatang: Let us use the `perform.DockerPull` function to do this via the API (which will be more controllable for us than using os/exec) its [here](https://github.com/eris-ltd/eris-cli/blob/develop/services/operate.go#L137).  Basically just need something like:

```go
srv := definitions.BlankService()
srv.Image = img
ops := definitions.BlankOperation()
err := perform.DockerPull(srv, ops)
```

compleatang:
This is ready once you are able to make that one change noted above. Feel free to make that and then pull it in. :+1:

util/init.go
zramsay: perform imports util in a non-trival way...this won't fly

zramsay:
should we silence the failing test on circle (see #containers)? other than that I think it's ready to go

PR #192, opened 2015-08-31
Exec is back

zramsay:
tests 
- pass on linux box, 
- pass then fail after package level tests on circle: `Get http:///var/run/docker.sock/v1.20/version: dial unix /var/run/docker.sock: no such file or directory.`
- chains_test fails on OSX because `index.docker.io` is down


compleatang:
nice. am gonna merge this to reconcile everything and then figure out where we are on the tests.

PR #194, opened 2015-08-31
fix & improve `eris init`, alias to `update`

### LINKED ISSUE #189: `eris update`
update the tool itself. 

can add build from source if Go installed now. binary updates will be blocking until that pipeline is fully completed.

zramsay:
closes #189, has a little of #203 
cleaning:
- [x] debug statements
- [x] deduplicate error messages & use some `logger.Debug`

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -1,18 +1,54 @@
 package commands
 
 import (
-	"github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/spf13/cobra"
 	ini "github.com/eris-ltd/eris-cli/initialize"
+	"github.com/eris-ltd/eris-cli/util"
+
+	"github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/spf13/cobra"
 )
 
 // flags to add: --no-clone
 var Init = &cobra.Command{
-	Use:   "init",
-	Short: "Initialize the ~/.eris directory with some default services and actions",
+	Use:     "init",
+	Aliases: []string{"update"},
+	Short:   "Initialize the ~/.eris directory with default files or update to latest version",
 	Long: `Create the ~/.eris directory with actions and services subfolders
 and clone eris-ltd/eris-actions eris-ltd/eris-services into them, respectively.
 `,
 	Run: func(cmd *cobra.Command, args []string) {
-		ini.Initialize(do.Pull)
+		Router(cmd, args)
 	},
+}
+
+func buildInitCommand() {
+	addInitFlags()
+}
+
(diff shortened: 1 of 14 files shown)


commands/update.go
```diff
@@ -0,0 +1,87 @@
...
+	out3, err := exec.Command("go", goArgs...).CombinedOutput()
+	if err != nil {
+		fmt.Printf("error with go install ./cmd/eris:\n%s\n", string(out3))
+		os.Exit(1)
+	}
+
+	fmt.Printf("Marmot update successful. Eris CLI Version is now: %s\n", VERSION)
+}
+
+func checkGitAndGo() {
```
compleatang: let's pull this function and the next into `utils` we'll likely need them for other things down the road.

commands/update.go
```diff
@@ -0,0 +1,87 @@
...
+	changeDirectory()
+
+	var branch string
+	if do.Branch != "" {
+		branch = do.Branch
+	} else {
+		branch = "master"
+	}
+
+	checkoutArgs := []string{"checkout", branch}
```
compleatang: may make sense to pull most of this function into utils.

commands/update.go
```diff
@@ -0,0 +1,87 @@
...
+	fmt.Printf("%s pulled successfully\n", branch)
+
+	goArgs := []string{"install", "./cmd/eris"}
+	out3, err := exec.Command("go", goArgs...).CombinedOutput()
+	if err != nil {
+		fmt.Printf("error with go install ./cmd/eris:\n%s\n", string(out3))
+		os.Exit(1)
+	}
+
+	fmt.Printf("Marmot update successful. Eris CLI Version is now: %s\n", VERSION)
```
compleatang: run `eris init`...? (i can see args either way)

**Iff** we do run `eris init` then we'd want to add a few flags so:

* `--tool` would only update the tool (current functionality)
* `--services` would only update the default services (just the pullRepo("eris-services")) basically
* `--actions` would only update the default actions ^
* `--all` (default) would do all the above.

commands/init.go
```diff
@@ -1,18 +1,59 @@
...
+	} else if !do.Tool {
+		do.Services = false
+		do.Actions = false
+		ini.Initialize(do)
+	}
+
+	if do.Tool {
+		util.UpdateEris(do.Branch)
+	}
+}
```
zramsay: can't seem to nail the logic down here. maybe my approach is wrong?

compleatang:
OK. I think that logic works well now. I've refactored it a bit to make it cleaner.

PR #198, opened 2015-09-01
for purifying common/go/ipfs

zramsay:


compleatang:
this should be ready, no?

PR #199, opened 2015-09-01
--env, plop

ebuchman:


ebuchman:
this needs some test fixes before merge

commands/chains.go
```diff
@@ -345,16 +355,19 @@ func addChainsFlags() {
 	chainsNew.PersistentFlags().StringVarP(&do.CSV, "csv", "", "", "render a genesis.json from a csv file")
 	chainsNew.PersistentFlags().StringVarP(&do.Priv, "priv", "", "", "pass in a priv_validator.json file (dev-only!)")
 	chainsNew.PersistentFlags().UintVarP(&do.N, "N", "", 1, "make a new genesis.json with this many validators and create data containers for each")
-	chainsNew.PersistentFlags().BoolVarP(&do.Run, "run", "r", false, "run the chain after creating")
+	chainsNew.PersistentFlags().BoolVarP(&do.Operations.PublishAllPorts, "publish", "p", false, "publish all ports")
+	chainsNew.PersistentFlags().StringSliceVarP(&do.Env, "env", "e", nil, "<key>=<value> pairs to pass into the container's bash env")
```
compleatang: "multiple env vars can be passed using the KEY1=val1,KEY2=val1 syntax"

commands/chains.go
```diff
@@ -461,6 +475,15 @@ func CurrentChain() {
 	IfExit(chns.CurrentChain(do))
 }
 
+func PlopChain(args []string) {
+	if len(args) != 2 {
```
compleatang: can use the args checker here:

```go
IfExit(ArgCheck(2, "eq", cmd, args))
```

definitions/do.go
```diff
@@ -20,6 +20,7 @@ type Do struct {
 	Lines         int      `mapstructure:"," json:"," yaml:"," toml:","` // XXX: for tail and logs
 	Timeout       uint     `mapstructure:"," json:"," yaml:"," toml:","`
 	N             uint     `mapstructure:"," json:"," yaml:"," toml:","`
+	Address       string   `mapstructure:"," json:"," yaml:"," toml:","`
```
compleatang: should we be worried about ballooning of the do struct? or just keep it big and use appropriately? i mean eris isn't a program we typically need to worry about memory bloat or things like that so probably fine.

services/operate.go
```diff
@@ -42,6 +42,11 @@ func StartService(do *definitions.Do) (err error) {
 		logger.Debugln("\t", s.Name, s.ServiceDeps, s.Service.Links, s.Service.VolumesFrom)
 	}
 
+	// NOTE: the top level service should be at the end of the list
```
compleatang: meant to mention that yesterday. this will help resolve any "container does not exist; cannot link to it" errors

definitions/do.go
zramsay: i wonder about this too ... just added three more flag vars myself

ebuchman: Ya so, I'm thinking pretty soon we might want to bust it into smaller chunks and have an args struct type for each (or most) of those main functions that currently take a Do

compleatang: let's move this here: https://github.com/eris-ltd/eris-cli/issues/202

compleatang:
I have no comments here. Think we should just merge these branches and fix the tests post merge cause they're all failing on the same thing.

PR #210, opened 2015-09-02
Ports, links

ebuchman:
- fixes csv handling on new so we can pass in two csvs (one for validators, one for accounts)
- add command for printing port mappings
- allow chains and services to link to arbitrary containers from command line (mostly to facilitate dev)

### DIFF
--- a/chains/operate.go
+++ b/chains/operate.go
@@ -15,6 +15,7 @@ import (
 	"github.com/eris-ltd/eris-cli/definitions"
 	"github.com/eris-ltd/eris-cli/loaders"
 	"github.com/eris-ltd/eris-cli/perform"
+	"github.com/eris-ltd/eris-cli/services"
 	"github.com/eris-ltd/eris-cli/util"
 
 	. "github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/eris-ltd/common/go/common"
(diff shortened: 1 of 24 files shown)


compleatang:
GTM?

chains/operate.go
```diff
@@ -214,19 +215,36 @@ func setupChain(do *definitions.Do, cmd string) (err error) {
 		return fmt.Errorf("Error making data directory: %v", err)
 	}
 
-	var csvFile, csvPath string
+	// we accept two csvs: one for validators, one for accounts
```
compleatang: did not know that!

chains/operate.go
```diff
@@ -280,16 +298,16 @@ func setupChain(do *definitions.Do, cmd string) (err error) {
 	envVars := []string{
 		fmt.Sprintf("CHAIN_ID=%s", do.ChainID),
 		fmt.Sprintf("CONTAINER_NAME=%s", containerName),
-		fmt.Sprintf("CSV=%v", csvPath),                 // for mintgen
-		fmt.Sprintf("CONFIG_OPTS=%s", configOpts),      // config.toml
-		fmt.Sprintf("REGISTER_ADDRESS=%s", do.Address), // use to sign registration txs for etcb
+		fmt.Sprintf("CSV=%v", csvPaths),               // for mintgen
+		fmt.Sprintf("CONFIG_OPTS=%s", configOpts),     // config.toml
+		fmt.Sprintf("REGISTER_PUBKEY=%s", do.Address), // use to sign registration txs for etcb
```
compleatang: am wondering now that we can give the edb container genesis json via env vars if we just do that rather than the current importing via the data dir paradigm....?

ebuchman: so instead of copying the file above, we just pass it in with the env vars here. im cool with that. is doing that with the priv key an issue? also, it means getting a list of the containers env variables will be cluttered with the contents of those files. is there some reason to prefer one over the other, i guess other than some of the complexity of copying into data containers (we'd still have to support that for the `--dir` flag)

chains/operate.go
ebuchman: new feature :)

ebuchman:
ya im GTM. i should note the `--links` are for raw container names, eg `--links=eris_services_whatever_1`. It's a dev feature.

compleatang:
sweet. let me finish reviewing then will merge.

chains/operate.go
compleatang: no special reason. i needed to add them for cloud boots of edb. happy to continue with current paradigm, but if we ever run into issues we could revisit at that time.

compleatang:
LGTM.

PR #211, opened 2015-09-02
dependencies refactor, chains dependencies

ebuchman:
dependencies now show up as

```
[dependencies]
services = [ "..." ]
chains = [ "..." ]
```

The current state of this PR does not use the chains dependencies from services (we should probably leave that for a much later TODO).

Services depending on services is as before.

Chains depending on services only works for simple services (ie. without recursive dependencies) - but we can fix that easily in this PR

Chains depending on chains require those chains to already be up and running else the command will fail.


compleatang:
good to merge or WIP?

compleatang:
K. this triple PR LGTM. will merge in a sec.

Issue #34, opened 2015-07-01
build namespaced genesis and config stash within ~/.eris/blockchains


compleatang:
(useful for testing purposes as well as for baas type situations).

this should be globally configurable reference to a specific file.

a base specific file should be compiled into the binary (like we do with erisdb and ipfs service definition files).

perhaps a default genesis.json repo (to mirror services and action repos) could be pulled in during Initialize.

dennismckinnon:
I like the idea of constructing a base file. since the genesis must have the account inserted though i doubt we can put these in a dedicated repo unless you just want the configuration settings copied. I think we probably can be more opinionated then that. someone can edit their default if they want but we should set the settings. fast_sync off for instance.

ebuchman:
i like the default behaviour being that it serves you a fresh genesis.json with the corresponding new key, so you can start running a one man chain right away. any predetermined genesis.json that correspondings to a running chain would be `fetch` (now known as `install`).

compleatang:
good point re the repo. we could use a `~/.eris/blockchains/genesis` directory and then have `caseythrowaway` in there. then we can use `eris chains new --genesis caseythrowaway asdf1234` that would use `~/.eris/blockchains/configs/default.toml` as the config.toml and `~/.eris/blockchains/genesis/caseythrowaway` as the genesis.json and spin up the container; alternatively `eris chains new asdf124` would use `~/.eris/blockchains/genesis/default` as the default (fresh) genesis.json.  

note, none of this would conflict with either of your points (I don't think)

ebuchman:
so i think the only thing this buys us is determinism re your address for testing contract systems that require hardcoded addresses (which is bad design anyways)

but it also buys a standard spot for genesis.json files, which is great, because we can cache them there too as we fetch em from etcb in our various adventures.

however I still like the default behaviour (no genesis specified) to be a fully fresh one with a new key and no default persistence (besides the data container).

compleatang:
OK. I'm happy with that default.

To be clear, I think we should have a standard spot for both genesis and config's although I think configs probably will just be the default....

ebuchman:
```
.eris/
  config/
    ipfs/
       ipfs.conf
    mint/
        config.toml
    ...
```
?

compleatang:
i was thinking more like:

```
.eris/
  blockchains/
    chainA.toml
    chainB.toml
    genesis/
       throwaway1.json
       composing.json
       composed2.json
   config/
       default.toml
```

[...]

PR #228, opened 2015-09-08
`chains new [name] --dir [name]` if [name] is in /config, #34

zramsay:


ebuchman:
I recently changed `mintgen` to create folders in `.eris/blockchains`. Not sure I see the need to use `.eris/blockchains/config` instead. @compleatang ?

chains/operate.go
```diff
@@ -207,6 +207,24 @@ func setupChain(do *definitions.Do, cmd string) (err error) {
 		do.ChainID = do.Name
 	}
 
+	//check in given path else check in config
```
ebuchman: `if given path does not exist, see if its a reference to something in ~/.eris/blockchains/config` 

be explicit in your comments!

chains/operate.go
```diff
@@ -207,6 +207,24 @@ func setupChain(do *definitions.Do, cmd string) (err error) {
...
+	if do.Path != "" {
+		src, err := os.Stat(do.Path)
+		if err != nil {
+			logger.Printf("path: %s does not exist, trying $HOME/.eris/blockchains/config/%s\n", do.Path, do.Name)
```
ebuchman: use Info or Debug. Print is only for things that should be printed no matter what ...

ebuchman: also this is weird because it totally ignores `do.Path`.

so I can do `eris chains new mexico --dir notmexico` and it will find the thing in mexico. hmmm

chains/operate.go
```diff
@@ -207,6 +207,24 @@ func setupChain(do *definitions.Do, cmd string) (err error) {
...
+			do.Path, err = util.PathChecker("chain", do.Name)
+			if err != nil {
+				return err
+			}
+		} else if !src.IsDir() {
```
ebuchman: combine this if statement into the above with an `||`

util/paths.go
```diff
@@ -0,0 +1,26 @@
+package util
+
+import (
+	"os"
+	"path/filepath"
+
+	. "github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/eris-ltd/common/go/common"
+)
+
+func PathChecker(typ, name string) (string, error) {
```
ebuchman: this is a nasty function. instead of taking `typ`, it should just take a path ...

zramsay: it takes typ so that it can be used for other parts of the stack in the future with a switch statement

ebuchman: but it currently doesnt support that. for now it should just take the path so its super clear

compleatang:
@ebuchman I think those directories are fine cause those will be in containers but the config name would be on the host so doubt there will be any conflicts. We *could* structure ecm to use this paradigm, but before we did that let's see if there are any challenges by first working in the host.

ebuchman:
so I should change mintgen to drop dirs in `blockchains/config` ?
not sure I understand your comment. I'm saying if I run mintgen on my host, itll drop a directory in eris/blockhains. would be great if that was good enough spot for me to use this new short form `--dir` feature. the only other thing in `blockchains/` on the host is some .toml files, and that will be the location of the dir in the container too

compleatang:
sorry. i misunderstood. seems logical that we'd connect these things to me. @zramsay any problems from your side? 

(its just removing the `config` from the default and from PathAnalyzer)

Issue #220, opened 2015-09-05
Remove Ensure Running


ebuchman:
https://github.com/eris-ltd/eris-cli/commit/4ebec25f591daca9852848eb7587beddd02df292#diff-bedfcf59ccad778eca6631b2cd237ca9R41

is a mess

We cannot do this in general. We should have *only* `EnsureIPFSRunning` and it should *only* send an http post to the gateway port.

compleatang:
Don't disagree with this necessarily. We may also want `EnsureKeysRunning` ....? 

@zramsay any thoughts?

zramsay:
`EnsureRunning` should just check that the service is running and if not, prompt user to start the service, but not actually start it for them. (while starting it for them is kinda nice, it creates too much abstraction...e.g. if `eris files put someFile` *always* starts IPFS (as it currently does), one may never need to run `eris services start ipfs`. I think that's bad

zramsay:
#229

PR #229, opened 2015-09-08
Dont ensure running, #220

zramsay:


files/files_test.go
```diff
@@ -122,6 +123,12 @@ func testsInit() error {
 	ifExit(err)
 	f.Write([]byte(content))
 
+	do1 := definitions.NowDo()
+	do1.Args = []string{"ipfs"}
+	err = services.StartService(do1)
+	time.Sleep(5 * time.Second)
```
ebuchman: sleep after checking the error :)

perform/docker_run.go
```diff
@@ -265,14 +265,14 @@ func DockerRun(srv *def.Service, ops *def.Operation) (string, error) {
...
-			return "", err
+			return err
 		}
 
 	} else {
 		logger.Infof("Successfully started service =>\t%s\n", srv.Name)
 	}
 
-	return id_main, nil
+	return nil
```
ebuchman: is there some reason not to return this? presumably it was only used by ensure running? i can see it being useful for other things potentially

services/load.go
```diff
@@ -22,44 +20,11 @@ func EnsureRunning(do *definitions.Do) error {
...
-					time.Sleep(500 * time.Millisecond)
-				}
-			}
-			time.Sleep(500 * time.Millisecond)
-			break MAIN_LOOP
-		}
+		errmsg := fmt.Sprintf("%s is not running, start it with `eris services start %s`\n", strings.ToUpper(do.Name), do.Name)
+		err := errors.New(errmsg)
+		return err
+		os.Exit(1)
```
ebuchman: can't have this. remember these functions need to work from an http server as well. you should define an error type called `ErrServiceNotRunning` and return that. 

actually, you're returning first so the Exit shouldn't trigger, but still, remove it and use a defined error.

ebuchman: Also fix the comments above EnsureRunning

perform/docker_run.go
zramsay: it never did return it until I needed the container ID to get the exposed ports for ping (which this PR is eliminating). it also made `DockerRun()` ugly

zramsay: we can revisit later if need be...

compleatang: think cleaner without the return (for now). when we get to agents and remotes we may want to get this info as a return value but let's get a better grasp on things at that time.

compleatang:
OK. This LGTM. @ebuchman? @zramsay you ready for me to merge?

PR #241, opened 2015-09-14
chains new should read defaults

compleatang:
the default config directory should be read when chains new is used and when no --dir flag has been given. the default chain files will first be copied into ~/.eris/data/BLAH/blockchains/BLAH then any individual file flags will be copied into the same location. This is the correct behaviour if for example, a user wants to have their default config.toml used but alongside a genesis.json which is not the default.


compleatang:
cc @ebuchman just give me a quick eye here to make sure you're ok with the proposed changes before i merge.

PR #242, opened 2015-09-14
godeps updates to major client libraries

compleatang:
* cobra (man page generator and did you mean support)
* pflag
* go-dockerclient (massages throughout mostly)


compleatang:
@zramsay @ebuchman note that we will need to fix some things the next time we update go-dockerclient. I forwarded past this commit and logs was failing. So until we get some time open up to fix that, let's leave it on this commit.

PR #244, opened 2015-09-16
UploadToContainer / docker cp; closes #131

### LINKED ISSUE #131: update data import/export to use Docker API functionality rather than current hack
use new docker API support for this feature (with backwards compatibility)...

depends upon: https://github.com/fsouza/go-dockerclient/issues/352 being successful closed.

zramsay:
the updated godep breaks `TestLogsService()`, despite the fact that there's no diff on the `DockerClient.Logs()`


compleatang:
Yup. I've noticed this as well. They've changed the writers somehow but it is unclear from their code what that has entailed. Until we have time to clean that this is gonna have to be put on hold.

zramsay:
this *might* be on our end. the last part of `TestChainNewDirGen()` fails for various reasons depending on a host of minor tweaks (some tracked in `docker_cp_test`). I tried with cherry-picking the commits from `UploadToContainer()` and it still fails. So it might be the function or it might be how we're calling it ...?

compleatang:
not inconceivable. @pietv when you finish the small writeup you are working on and labels see if you can take a look. Zach can fill you in on any background you may need.

zramsay:
branched from `contractsTests`; ready to go when that branch is

Issue #66, opened 2015-07-05
squash echo back of commands when in `exec --interactive`


compleatang:
![screenshot from 2015-07-05 13 41 08](https://cloud.githubusercontent.com/assets/590037/8511526/d508d3a6-231b-11e5-9bca-c9dc249d73ab.png)

compleatang:
I still have no idea how to do this.

PR #275, opened 2015-10-04
This PR fixes the echo issue with interactive attach

pietv:
* Fixes the echo issue #66 with commands that attach to containers interactively (such as `data exec`) by putting the terminal into raw mode before attach and restoring the terminal's previous state upon exit / interrupt.
* Adds the external package `"github.com/docker/docker/pkg/term"` to handle the terminal settings portably. Godeps updated.
* Allows processing of some control chars (`^D`, `^L`, `^A`, etc.) 

image


compleatang:
:+1:

Issue #217, opened 2015-09-05
install script


compleatang:
- [ ] finalize
  * wrap docker's script
  * ensure its linux
  * install pkgs (depends on rpm publication)
  * run `eris ini`
- [ ] deploy to https://get.eris.industries

compleatang:
depends on #186

compleatang:
moving this to 0.12 need to more thoroughly test yum.

compleatang:
closing for now common/cloud/setup will suffice

Issue #186, opened 2015-08-30
figure out how to build for yum


compleatang:
https://github.com/alanfranz/docker-rpm-builder

pietv:
Building for YUM involves building RPMs and making them available via an HTTP server. In a few steps:

## 1. Generate a PGP key for signing RPMs

Generate a PGP key pair and make a public key available via the HTTP server:

```
$ gpg2 --gen-key

$ gpg2 --armor --output /var/www/html/eris/RPM-GPG-KEY --export 'Eris Industries'
```

## 2. Build and sign RPM packages

The `eris` name for the package appears to be already taken (a "Client-side session layer for Atlas-C++"). We can choose the name later, but `eris-tools` could do for now:

```
$ rpmbuild -ba --sign eris-tools.spec

$ cp RPMS/x86_64/eris/eris-tools-0.10.4-1.x86_64.rpm /var/www/html/eris

$ cp SRPMS/eris-tools-0.10.4-1.src.rpm /var/www/html/eris/source/

```

The current `eris-tools` package is functional but missing a few important pieces: `README`, `COPYING` files, man page(s), and possibly completion and `ChangeLog` files.

## 3. Index the repo directories

```
$ cd /var/www/html/eris

$ createrepo x86_64
Spawning worker 0 with 1 pkgs
Workers Finished
Saving Primary metadata
Saving file lists metadata
Saving other metadata
Generating sqlite DBs
Sqlite DBs complete

$ createrepo sources
Spawning worker 0 with 1 pkgs
Workers Finished
...
```

## 4. Test the packages

Place this `eris.repo` file below under `/etc/yum.repos.d` directory. (The http://188.166.124.29:8100/ is a sample temporary server.)

```ini
[eris]
name=Eris
baseurl=http://188.166.124.29:8100/eris/x86_64/
metadata_expire=1m
enabled=1
gpgkey=http://188.166.124.29:8100/eris/RPM-GPG-KEY
gpgcheck=1

[eris-source]
name=Eris Source
baseurl=http://188.166.124.29:8100/eris/source/
metadata_expire=1m
enabled=1
gpgkey=http://188.166.124.29:8100/eris/RPM-GPG-KEY
gpgcheck=1
```

### Fedora 22

```
[root@fedora22 yum.repos.d]# rpm -qa \*gpg-pubkey\*
gpg-pubkey-8e1431d5-53bcbac7


[root@fedora22 yum.repos.d]# dnf repolist
[...]

Issue #185, opened 2015-08-30
generate man pages


compleatang:
https://github.com/spf13/cobra/commit/9e7788657f7d4cdc20007a16ba586b4ba0c3aab1

Issue #39, opened 2015-07-03
[Core] Build shell integration 


compleatang:
- [ ] bash completions.
- [ ] zsh package 

Namespace aliasing... `mintx` routing to `eris actions mintx do`...  Or to `docker run --volumes-from $(eris projects inspect data_container) eris/mintx` whatever. Need to consider flag conversion.

compleatang:
http://stackoverflow.com/questions/3249432/i-have-a-bash-tab-completion-script-is-there-a-simple-way-to-use-it-from-zsh

compleatang:
this isn't gonna happen just now. moving to 0.11

compleatang:
zsh completions issue on dep lib -> https://github.com/spf13/cobra/issues/107

koryk:
Was able to automatically generate a bash completion file by calling the  GenBashCompletionFile method on the ErisCmd object - https://godoc.org/github.com/spf13/cobra#Command.GenBashCompletionFile

The resulting file can be found here
https://gist.github.com/koryk/9b0625509def80cdefa451b61e8c9d25

pietv:
Awesome! Thanks very much for this! :+1: Will play with this a little bit and integrate to Linux and Mac package generation. Automatic completion is so much better than the manual one; totally missed this Cobra functionality.

zramsay:
obsolete

PR #286, opened 2015-10-12
RPM spec file

pietv:
* RPM spec file to be tested with issues #217, #186, #185, and #39.


tests/hack/eris-tools.spec
```diff
@@ -0,0 +1,35 @@
+%{define} erisbuilddir /var/tmp/eris-rpmbuild.tmp
+
+Summary: Eris is a platform for building, testing, maintaining, and operating distributed applications with a blockchain backend. Eris makes it easy and simple to wrangle the dragons of smart contract blockchains.
+Name: eris-tools
+Version: 0.10.4
```
compleatang: @pietv is it possible to autopopulate this somehow so that we don't have to remember to update it whenever we do a version bump?

compleatang:
This looks good save for that one comment above. If there is a way that we can autopopulate that either with rpmbuild flag or some other way (maybe in the tests/hack/release.sh script?) it would be ideal. just let me know if it isn't possible and we'll add a check in the `tests/hack/release.sh` script so that future releases will get flagged to check that like and make sure it is updated.

pietv:
Yeah, it's perfectly possible to populate the version and build numbers through environment variables:

```
$ ERIS_VERSION=0.10.4 ERIS_RELEASE=1 rpmbuild -ba --sign eris-tools.spec
```

The `ERIS_RELEASE` variable could be set to a build number or be incremental (by definition, it's the number of times the RPM package has been rebuilt with rpmbuild); or it could be always 1.

compleatang:
This is great, only holding off on merging because we don't need to worry about merge conflicts and also I will want to update my release scripts significantly here before we release 0.10.4 (middle next week...?)

Issue #258, opened 2015-09-19
Move from current container name parsing paradigm to using docker labels paradigm


compleatang:
To track:

- `SWARM`
- `MACHINE`
- `USER` (we have from git info)
- `TYPE`
- `TESTS`
- `NAME`
- `SERVICE_ID`
- `DATA_CONTAINER_ID`

compleatang:
relates to #231 #170

pietv:
@compleatang @zramsay @ebuchman  a few quick questions:

* What information would the `TESTS` label hold? 

* Why both `SERVICE_ID` and `DATA_CONTAINER_ID` and not just `ID` and `TYPE`?

* After the schema change, the container names still need to be unique, does anyone oppose using timestamps alongside eris prefix (something like: `eris-20150930-194054`). The alternatives could be using the standard Docker naming convention (person name+adjective), random numbers, sequential numbers, unix epoch (seconds).

compleatang:
> What information would the TESTS label hold?

We should use a `TEST` label which should just be `true` or `false`, along with a `TEST_ID` field which can be a UUID for a specific run of the test suite (which is not currently used, but moving in this direction would allow that to happen).

> Why both SERVICE_ID and DATA_CONTAINER_ID and not just ID and TYPE?

Excellent idea. 

> Unique container names

I'm easy. Not sure what @ebuchman thinks, but I'd be in favor of `ipfs-UUID` or `ipfs-TIMESTAMP` paradigm. We probably don't need the eris prefix necessarily, although it may help users who move between docker porcelain and eris plumbing to have an idea of which is which. 

One thing we should add to the labeling is a `CONTAINER_NUMBER` field. This will help Ethan in his testing of multiple chain containers on he same machine to be able to track those.

pietv:
@compleatang thanks! I like the `ipfs-UUID` bertter than `ipfs-TIMESTAMP`. The UUID is much longer than the timestamp, but it can be truncated in `eris ls` (Not in `docker ps`, though, but that's not a big issue.)

Current labels:

- `SWARM`
- `MACHINE`
- `USER` (we have from git info)
- `ID`
- `NAME`
- `TYPE`
- `SERVICE`
- `CONTAINER_NUMBER`
- `TEST`
- `TEST_ID`

compleatang:
With the Go UUID lib you can truncate in the `-` so it becomes like 6 or 8 chars or something.

[...]

PR #302, opened 2015-10-25
Container labels refactoring

pietv:
Related to #258. This PR contains refactoring of code to accomodate container labels. The old naming scheme is still used alongside the new one, that is the labels can be observed with `docker inspect` and checked against the container name.

I decided to push this potentially dangerous change first to be sure that if anything breaks, it's not because of labels.

The remaining bit will be isolated in a few `util/container*.go` files.

The one difference with the requirements is that the `SERVICE` label of a volumes-from data container contains the name and not the ID of the service (the service itself may or may not be running to determine ID).

```
        "MacAddress": "",
        "OnBuild": null,
        "Labels": {
            "eris:CONTAINER_NUMBER": "1",
            "eris:ERIS": "true",
            "eris:NAME": "tinydns",
            "eris:SERVICE": "eris_service_tinydns_1",
            "eris:TYPE": "data",
            "eris:USER": "Peter-Vypov"
        }
    }

```

The `ID`, `MACHINE`, and `SWARM` labels are not stored in the container and will be determined at its run time.

The `TEST` and `TEST_ID` labels can be added either to `util.Labels()`, `util.SetLabel()` functions or be determined at run time, but they're not added anywhere yet.

Changes:

* A few Docker* function signatures has changed to avoid adding a 5th or 6th parameter to them. The parameters are now added to `Operations` or moved from `Do`: ContainerType, Args, Interactive, and Volume:

```
DockerCreateDataContainer(ops)
DockerRunVolumesFromContainer(ops, service)
DockerRunInteractive(srv, ops)
DockerRename(ops, newName)
```
* `DockerRename()` function now removes and recreates containers, not renames them. This is to support labels, which cannot be modified at container run time. This has one side effect that the changes made to the container before rename are not retained. This required to rearrange the `TestDataRename()` test to put it after the import and export tests. (The other side effect is that this is a longer operation.)
* `loaders.LoadDataDefinition()` which doesn't use its own type. It populates the `Operation` with initial settings and labels.
* `util.OverWriteOperations()` rewrite to allow using other structs.

Locally this has been tested with Docker 1.5 and 1.9 (on Debian Jessie, Fedora 22, and El Capitan 10.11.1 Beta).

### DIFF
--- a/chains/manage.go
+++ b/chains/manage.go
@@ -265,8 +265,10 @@ func RenameChain(do *definitions.Do) error {
 		}
 
 		chainDef.Name = newNameBase
-		// [pv]: why are these emptied at rename?
+		// Generally we won't want to use Service.Name
+		// as it will be confused with the Name.
 		chainDef.Service.Name = ""
+		// Service.Image should be taken from the default.toml.
 		chainDef.Service.Image = ""
 		err = WriteChainDefinitionFile(chainDef, newFile)
 		if err != nil {
--- a/chains/manage.go
+++ b/chains/manage.go
@@ -41,11 +41,11 @@ func RegisterChain(do *definitions.Do) error {
 
 	// set chainid and other vars
 	envVars := []string{
-		fmt.Sprintf("CHAIN_ID=%s", do.ChainID),      // of the etcb chain
-		fmt.Sprintf("PUBKEY=%s", do.Pubkey),         // pubkey to register chain with
-		fmt.Sprintf("ETCB_CHAIN_ID=%s", etcbChain),  // chain id of the etcb chain
-		fmt.Sprintf("NODE_ADDR=%s", do.Gateway),     // etcb node to send the register tx to
-		fmt.Sprintf("NEW_P2P_SEEDS=%s", do.Args[0]), // seeds to register for the chain // TODO: deal with multi seed (needs support in tendermint)
+		fmt.Sprintf("CHAIN_ID=%s", do.ChainID),                 // of the etcb chain
+		fmt.Sprintf("PUBKEY=%s", do.Pubkey),                    // pubkey to register chain with
+		fmt.Sprintf("ETCB_CHAIN_ID=%s", etcbChain),             // chain id of the etcb chain
+		fmt.Sprintf("NODE_ADDR=%s", do.Gateway),                // etcb node to send the register tx to
+		fmt.Sprintf("NEW_P2P_SEEDS=%s", do.Operations.Args[0]), // seeds to register for the chain // TODO: deal with multi seed (needs support in tendermint)
 	}
 	envVars = append(envVars, do.Env...)
 
@@ -60,9 +60,11 @@ func RegisterChain(do *definitions.Do) error {
 
 	logger.Debugf("Starting chain container via Docker =>\t%s\n", chain.Service.Name)
 	logger.Debugf("\twith Image =>\t\t%s\n", chain.Service.Image)
-	cmd := []string{loaders.ErisChainRegister}
-	dataContainerName := util.DataContainersName(chain.Name, chain.Operations.ContainerNumber)
-	_, err = perform.DockerRunVolumesFromContainer(dataContainerName, false, cmd, chain.Service)
+	chain.Operations = loaders.LoadDataDefinition(chain.Service.Name, do.Operations.ContainerNumber)
+	chain.Operations.Interactive = do.Operations.Interactive
+	chain.Operations.Args = []string{loaders.ErisChainRegister}
+
+	_, err = perform.DockerRunVolumesFromContainer(chain.Operations, chain.Service)
 
 	return err
 }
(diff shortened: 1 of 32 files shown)


pietv:
@compleatang, @ebuchman what do you think?

chains/manage.go
```diff
@@ -263,6 +265,7 @@ func RenameChain(do *definitions.Do) error {
 		}
 
 		chainDef.Name = newNameBase
+		// [pv]: why are these emptied at rename?
```
compleatang: We pull Service.Image from the default.toml is the theory. This is because generally we won't want to have lagging erisdb containers (lagging meaning disjointed from the cli version) unless the users expressly request it. So the general state of a chain definition file would be to have these fields be empty.. Generally we won't want to use `Service.Name` as it will get confusing for folks (`name` is a top level function separate from `Service.Name` obviously).

chains/operate.go
```diff
@@ -135,7 +134,8 @@ func startChain(do *definitions.Do, exec bool) error {
 	logger.Debugf("\twith AllPortsPublshd =>\t%v\n", chain.Operations.PublishAllPorts)
 
 	if exec {
-		err = perform.DockerRunInteractive(chain.Service, chain.Operations, do.Args, do.Interactive, "")
+		chain.Operations.Args = do.Operations.Args
+		err = perform.DockerRunInteractive(chain.Service, chain.Operations)
```
compleatang: I like this movement of `Args` into Operations struct. It cleans up the code a bunch.

perform/docker_run.go
```diff
@@ -23,21 +22,34 @@ import (
...
+var (
+	ErrContainerExists = errors.New("container exists")
+)
 
-	srv := def.BlankServiceDefinition()
-	srv.Operations.DataContainerName = util.DataContainersName(srvName, containerNumber)
-	optsData, err := configureDataContainer(srv.Service, srv.Operations, nil)
-	if err != nil {
-		return err
+// DockerCreateDataContainer creates a Docker container of type data specified
```
compleatang: Super! We've need to get some more goDocs style documentation so thanks for getting this ball rolling. :dragon_face:

perform/docker_run.go
```diff
@@ -343,7 +405,7 @@ func DockerRunInteractive(srv *def.Service, ops *def.Operation, args []string, i
 	}
 	id_main = servContCreated.ID
 
-	// Set terminal into row mode, and restore upon container exit.
+	// Set terminal into raw mode, and restore upon container exit.
```
compleatang: `row` mode please ;)

chains/manage.go
pietv: Great! Thanks for the detailed explanation. I see it now. There was a temptation really to set back the image.

perform/docker_run.go
pietv: Great, thanks! That was useful to remember which parameters are carried around.

perform/docker_run.go
pietv: :)

pietv: Or you mean for real? It's `raw` vs `cooked`?

compleatang: hahaha. no. was just making a joke! :)

pietv: Yay

chains/manage.go
compleatang: Yeah. Clearing the image I also go back and forth on but this is generally the paradigm we're using throughout the tool (clearing the image and using default until a chain is graduated into a service in which time the image is "sunk" into the service definition file).

util/container_operations.go
```diff
@@ -1,26 +1,86 @@
...
-	opsBase.PublishAllPorts = OverWriteBool(opsBase.PublishAllPorts, opsOver.PublishAllPorts)
+var (
+	ErrMergeParameters = errors.New("parameters are not pointers to struct")
+)
+
+// Merge replaces blank fields of base with those from over; also merges maps
+// and slices of both. Base and over are pointers to structs. The result is stored
+// in base. Merge returns ErrMergeParameters if either base or over are not
+// pointers to structs.
+func Merge(base, over interface{}) error {
```
compleatang: #killingit This is super helpful. We've need a proper reflect and merge function for a while.

compleatang:
All of this looks great to me. Gonna pull and run the test suite to see if I can see what the hell circle is doing and then I *think* it should be ready to merge.

ebuchman:
this is a beautiful PR. thanks Peter!

ebuchman:
@pietv apologies that we just merged a few things and this now needs a rebase. Looks like you got it passing tho yay! Want to rebase so we can merge?

pietv:
@ebuchman  thanks for the heads up and no worries! I think I forgot to
remove the "don't merge" label myself. I'll rebase!

Yeah, it looks like the tests failed on Circle because they timeouted. After the timeout increase they pass. Yay!

Issue #225, opened 2015-09-07
command helpers & error messages


zramsay:
This issue is ongoing and is used for suggesting clarifications/modifications to command helpers for `eris` and all its subcommands.

If a command is confusing, please ask for clarification here. Same goes for unclear error messages
- [ ] [zr] update error/debug messages when commands change. e.g. #231

compleatang:
closing this in favor of more specific issues now that we have done almost a full audit of these.

zramsay:
:+1:

PR #317, opened 2015-11-11
eris man command

pietv:
`eris man` command to display or generate a man page.

```
$ eris man

$ eris man > eris.1
$ nroff -mdoc eris.1 | less
```

The command doesn't use the Cobra's `GenMan()` or `GenManTree()` commands and uses Go templates to populate man markup directly (for more flexibility and to avoid using intermediary Markdown format).

The are a couple of changes made to how the command line help output looks:

cli output

1. command line parameters are now capitalized to differentiate between command line itself and a parameter name. Square brackets now mean an optional parameter. 
2. examples are now prefixed with the `$` dollar sign so that they look better in the man page.

man output

This fixes the issue #185.

### DIFF
--- a/commands/actions.go
+++ b/commands/actions.go
@@ -17,8 +17,8 @@ import (
 // flags to add: --global --project
 var Actions = &cobra.Command{
 	Use:   "actions",
-	Short: "Manage and Perform Structured Actions.",
-	Long: `Display and Manage actions for various components of the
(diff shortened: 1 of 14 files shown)


zramsay:
+1 for more consistency with helpers; #225

compleatang:
:+1:

Issue #321, opened 2015-11-12
interactive `services exec` command doesn't restore the terminal state on exit


pietv:
The interactive `services exec` command leaves the terminal in a broken state on exit, which requires issuing the `reset` command.

PR #326, opened 2015-11-18
Fix the broken terminal state after exec commands

pietv:
The upgraded Docker client library closes the `stdin` of the attached container on exit. Use a proxy pipe for `stdin`.

Fixes the issue #321.


compleatang:
:+1: :dragon_face:

Issue #294, opened 2015-10-18
expose eris-keys functions as `eris keys` 


compleatang:
- [x] gen
- [ ] lock
- [ ] unlock
- [ ] name
- [ ] ls (note this doesn't currently exist in eris-keys)
- [x] pub (what does this do....?) [zr] take --addr/name & returns pubkey
- [ ] import
- [x] export
- [x] [zr] `mintkey mint` -> `eris keys convert` or something

zramsay:
that's a lot of functions to expose - maybe limit it to just a few basic ones to begin with

compleatang:
agree. that was just meant as the exhaustive list of available commands. please do curate.

ebuchman:
Close?

zramsay:
ye

PR #333, opened 2015-11-20
eris keys

zramsay:
related to #294, useful for doing simple things with cli w/o using exec:
`eris keys gen`
`eris keys pub ADDR`
`eris keys convert ADDR` -> mintkey mint ADDR
`eris keys export`
`eris chains make-genesis NAME PUB`

### DIFF
--- a/commands/chains.go
+++ b/commands/chains.go
@@ -348,8 +348,7 @@ var chainsMakeGenesis = &cobra.Command{
 	Short: "Generates a genesis file.",
 	Long: `Generates a genesis file with chainNAME and a single pubkey.
 
-Command is equivalent to:
-$ eris chains exec someChain "mintgen known NAME KEY"
+Command is equivalent to: [eris chains exec someChain "mintgen known NAME KEY"]
 
 but does not require a pre-existing chain to execute.
 
--- a/commands/chains.go
+++ b/commands/chains.go
@@ -5,6 +5,8 @@ import (
 	"strings"
 
 	chns "github.com/eris-ltd/eris-cli/chains"
+	def "github.com/eris-ltd/eris-cli/definitions"
+	srv "github.com/eris-ltd/eris-cli/services"
 	"github.com/eris-ltd/eris-cli/util"
 
 	. "github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/eris-ltd/common/go/common"
@@ -59,6 +61,7 @@ func buildChainsCommand() {
 	Chains.AddCommand(chainsUpdate)
 	Chains.AddCommand(chainsRemove)
 	Chains.AddCommand(chainsGraduate)
+	Chains.AddCommand(chainsMakeGenesis)
 	addChainsFlags()
 }
 
@@ -342,6 +345,21 @@ Command will cat local chains definition file.`,
 	Run: CatChain,
 }
 
+var chainsMakeGenesis = &cobra.Command{
+	Use:   "make-genesis NAME KEY",
+	Short: "Generates a genesis file.",
+	Long: `Generates a genesis file with chainNAME and a single pubkey.
+
+	Command is equivalent to:
+		$ eris chains exec someChain "mintgen known NAME [KEY]"
+
+	but does not require a pre-existing chain to execute.
+
+	see https://github.com/eris-ltd/mint-client 
+	for more info on the mint client tools`,
+	Run: MakeGenesisFile,
+}
+
 //----------------------------------------------------------------------
 
 func addChainsFlags() {
@@ -608,4 +626,43 @@ func CatChain(cmd *cobra.Command, args []string) {
 	IfExit(ArgCheck(1, "ge", cmd, args))
 	do.Name = args[0]
 	IfExit(chns.CatChain(do))
+}
+
+func MakeGenesisFile(cmd *cobra.Command, args []string) {
+	IfExit(ArgCheck(2, "ge", cmd, args))
+
+	//otherwise it'll start its own keys server that won't have the key needed...
+	do.Name = "keys"
+	IfExit(srv.EnsureRunning(do))
+
+	doThr := def.NowDo()
+	doThr.Chain.ChainType = "throwaway" //for teardown
+	doThr.Name = "default"
+	doThr.Operations.ContainerNumber = 1
+	doThr.Operations.PublishAllPorts = true
+
+	logger.Infof("Starting chain from MakeGenesisFile =>\t%s\n", doThr.Name)
+	if er := chns.StartChain(doThr); er != nil {
+		fmt.Printf("error starting chain %v\n", er)
+	}
+
+	chainName := strings.TrimSpace(args[0])
+	pubkey := strings.TrimSpace(args[1])
+
+	doThr.Operations.Args = []string{"mintgen", "known", chainName, fmt.Sprintf("--pub=%s", pubkey)}
+	doThr.Chain.Name = "default" //for teardown
+
+	// pipe this output to /chains/chainName/genesis.json
+	err := chns.ExecChain(doThr)
+	if err != nil {
+		fmt.Printf("exec chain err: %v\n", err)
+		do.Rm = true
+		do.RmD = true
+		IfExit(chns.CleanUp(doThr))
+	}
+
+	do.Rm = true
+	do.RmD = true
+	defer chns.CleanUp(doThr) // doesn't clean up keys but that's ~ ok b/c it's about to be used...
+
 }
--- a/commands/chains.go
+++ b/commands/chains.go
@@ -5,8 +5,6 @@ import (
 	"strings"
 
 	chns "github.com/eris-ltd/eris-cli/chains"
-	def "github.com/eris-ltd/eris-cli/definitions"
-	srv "github.com/eris-ltd/eris-cli/services"
 	"github.com/eris-ltd/eris-cli/util"
 
 	. "github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/eris-ltd/common/go/common"
@@ -350,13 +348,12 @@ var chainsMakeGenesis = &cobra.Command{
 	Short: "Generates a genesis file.",
 	Long: `Generates a genesis file with chainNAME and a single pubkey.
 
-	Command is equivalent to:
-		$ eris chains exec someChain "mintgen known NAME [KEY]"
+Command is equivalent to:
+$ eris chains exec someChain "mintgen known NAME KEY"
 
-	but does not require a pre-existing chain to execute.
+but does not require a pre-existing chain to execute.
 
-	see https://github.com/eris-ltd/mint-client 
-	for more info on the mint client tools`,
+see https://github.com/eris-ltd/mint-client for more info`,
 	Run: MakeGenesisFile,
 }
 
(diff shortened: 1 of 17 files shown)


pietv:
My couple of cents:
* Could the functionality of keys functions moved from `commands` directory to `keys` and `chains`? 
That way you don't need to put tests inside `commands`?

data/operate.go
```diff
@@ -150,7 +149,13 @@ func ExportData(do *definitions.Do) error {
...
 		// //   the temp contents there
-		prevDir := filepath.Join(DataContainersPath, do.Name)
+		var prevDir string
+		if do.Name == "keys" { //hack for `eris keys export`
+			prevDir = do.ErisPath
+		} else {
+			prevDir = filepath.Join(do.ErisPath, do.Name)
+		}
+
 		if _, err := os.Stat(prevDir); os.IsNotExist(err) {
```
pietv: * Could we get away without this hack (and `ErisPath` variable) maybe by using the parent dir automatically (something `filepath.Abs(filepath.Join(path, "..")`)?

commands/chains.go
```diff
@@ -342,6 +345,21 @@ Command will cat local chains definition file.`,
...
+		$ eris chains exec someChain "mintgen known NAME [KEY]"
+
+	but does not require a pre-existing chain to execute.
+
+	see https://github.com/eris-ltd/mint-client 
+	for more info on the mint client tools`,
+	Run: MakeGenesisFile,
+}
+
 //----------------------------------------------------------------------
```
pietv: * The `keys` and `chain` commands help / man output looks a bit different than for other commands:

cli output

Not really a big deal now, but needs to be changed some time, I think.

For comparison (`chains inspect`: https://github.com/eris-ltd/eris-cli/blob/develop/commands/chains.go#L257-L269 or `chains register`: https://github.com/eris-ltd/eris-cli/blob/develop/commands/chains.go#L96-L114)

The difference is: 1) no offset / tab at line start. 2) commands in `Long` are inlined and within [eris ...] or go into `Example` with a description. 3) A new line in `Long` should not start with [eris ...], better "The [eris ...]" (just a matter of ease on the eye in man). This seems to be the current acquired style. Can be changed to something else, but better be done for every command at once.

pietv:
LGTM! Great stuff! :shipit: :smiley_cat:

data/operate.go
compleatang: :+1: Peter's right this could be quite fragile....

commands/chains.go
compleatang: yeah. probably have to train VIM not to indent those lines by default (which is what would happen with most text editors), because when it renders those long strings the leading whitespace gets added. Let's get this fixed for harmony with the rest of the help screens before we merge this.

commands/chains.go
```diff
@@ -609,3 +627,42 @@ func CatChain(cmd *cobra.Command, args []string) {
...
+		do.Rm = true
+		do.RmD = true
+		IfExit(chns.CleanUp(doThr))
+	}
+
+	do.Rm = true
+	do.RmD = true
+	defer chns.CleanUp(doThr) // doesn't clean up keys but that's ~ ok b/c it's about to be used...
+
+}
```
compleatang: Let's move this functionality into the chains directory. We want to keep the commands package as lean on actual functionality as possible. All that should really be there is arg checking and then error handling after passing into the primary `do`ing function (which should be in chains pkg)

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+
+	It is for development only. Advanced functionality is available via:
+		$ eris services exec keys "eris-keys ..."
+		
+	see https://github.com/eris-ltd/eris-keys for more info.
+	
+	Mint tools are available via an erisdb container:
+		$ eris chains exec chainName "mint..."
+
+	see https://github.com/eris-ltd/mint-client for more info`,
```
compleatang: See comments above on leading whitespace as it applies here as well.

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+var Keys = &cobra.Command{
+	Use:   "keys",
+	Short: "Do specific tasks with keys *for dev only*.",
+	Long: `The keys subcommand is an opiniated wrapper around
+	eris-keys and requires a keys container to be running. 
+
+	It is for development only. Advanced functionality is available via:
+		$ eris services exec keys "eris-keys ..."
+		
+	see https://github.com/eris-ltd/eris-keys for more info.
```
compleatang: Docs are built => https://docs.erisindustries.com/documentation/eris-keys/ (let's route users there instead of GH; where our analytics don't reach).

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+var keysGen = &cobra.Command{
+	Use:   "gen",
+	Short: "Generates a key using the keys container.",
+	Long: `Generates a key using the keys container.
+
+	Key is saved in keys data container and can be exported to host with:
+		$ eris keys export
+
+	Command is equivalent to:
+		$ eris services exec keys "eris-keys gen --no-pass"`,
```
compleatang: leading whitespace ...

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+var keysExport = &cobra.Command{
+	Use:   "export",
+	Short: "Export all keys from container to host.",
+	Long: `Export all keys from container to host.
+	
+	Takes the contents of 
+		/home/eris/.eris/keys/data/ 
+
+	in the keys container and copies everything to
+		/home/user/.eris/keys/data/
```
compleatang: `$HOME/.eris/keys/data` <- will capture both linux and osx (and sorta windows)

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+
+	on the host.`,
+	Run: ExportKey,
+}
+
+var keysConvert = &cobra.Command{
+	Use:   "convert ADDR",
```
compleatang: let's add a required 2nd arg here of CHAINNAME. 

functionally it should be `eris keys convert 11111... idiaminchain` which should dump a priv_validator into ~/.eris/chains/idiaminchain/priv_validator.json ... since that's the primary use case let's do that. make sure to catch if the ~/.eris/chains/idiaminchain` directory doesn't exist and create it of course.

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+
+	Command is equivalent to:
+		$ eris services exec keys "eris-keys gen --no-pass"`,
+	Run: GenerateKey,
+}
+
+var keysPub = &cobra.Command{
+	Use:   "pub ADDR",
+	Short: "Returns a pubkey given an address.",
+	Long: `Returns a pubkey given an address.
```
compleatang: Returns a machine readable public key given an address.

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+	
+	Command is equivalent to:
+		$ eris services exec keys "eris-keys pub --addr ADDR"`,
+	Run: GetPubKey,
+}
+
+var keysExport = &cobra.Command{
+	Use:   "export",
```
compleatang: let's add an *optional* ADDR argument here to only export *one* addr. 

also let's add a `--dest` flag here so users can say where they wanna save the keys on the host.

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+
+	IfExit(ArgCheck(0, "eq", cmd, args)) // no args needed!
+
+	do.Name = "keys"
+	do.Operations.ContainerNumber = 1
+	IfExit(srv.EnsureRunning(do))
+	do.Operations.Interactive = false
+	do.Operations.Args = []string{"eris-keys", "gen", "--no-pass"}
+
+	IfExit(srv.ExecService(do))
```
compleatang: I know these are pretty slim functions, but I agree with Peter that I think they should be moved into a very slim keys directory/package. We are going to add more functionality here but over time we'll be adding more. We should generally stay away from anything more than arg checking and error checking on the backend within the commands package/directory.

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+		$ eris chains exec chainName "mint..."
+
+	see https://github.com/eris-ltd/mint-client for more info`,
+	Run: func(cmd *cobra.Command, args []string) { cmd.Help() },
+}
+
+func buildKeysCommand() {
+	Keys.AddCommand(keysGen)
+	Keys.AddCommand(keysPub)
+	Keys.AddCommand(keysExport)
```
compleatang: `Keys.AddCommand(keysImport)` -> if we're gonna have export, let's also have import which will be probably more useful than anything else. To do import we have to "cat" the file on the host and pipe that into eris-keys import (which can take raw json).

commands/keys.go
```diff
@@ -0,0 +1,160 @@
...
+	do.Name = "default"
+	tmp := do.Name
+	do.Operations.PublishAllPorts = true
+	IfExit(chains.ThrowAwayChain(do))
+	do.Chain.Name = do.Name // setting this for tear down purposes
+	logger.Debugf("ThrowAwayChain booted =>\t%s\n", do.Name)
+
+	do.Name = tmp
+	addr := strings.TrimSpace(args[0])
+	do.Operations.Args = []string{"mintkey", "mint", addr}
```
compleatang: `mintkey` binary is in eris/keys image so this functionality can mirror the other commands rather than worrying about the Throwaway chain which adds unnecessary complexity.

compleatang:
Mostly those should be minor changes. If there is anything in those comments that will require a significant time let me know.

commands/keys.go
zramsay: it's much cleaner to pipe it into the file...and keeps things simple. otherwise, there's things going on under the hood that aren't clear to the user...

commands/keys.go
zramsay: sgtm. in hindsight, they were put there just to get things up and running.

commands/keys.go
zramsay: :+1:

commands/keys.go
zramsay: @compleatang do we have the same for `mint-client`? don't see it anywhere on the docs site (used in chains)

commands/keys.go
zramsay: agree to make the helper more clear. not hugely in favour of adding *too* much functionality...users can exec.

commands/keys.go
compleatang: @zramsay we don't currently. given the upcoming edb refactor i wasn't sure what the cost benefit was on getting them in now, but all in all it is pretty easy.

commands/keys.go
compleatang: makes sense

commands/keys.go
zramsay: will do ADDR as flag...optional args can be confusing (especially if in `Use:`) 
`--dest` added & is (along with import) the reasoning behind the Import/Export Data paths refactor (which is now more versatile)

compleatang:
This looks great! Now to figure out why the tests don't pass...

services/load.go
```diff
@@ -21,7 +22,8 @@ func EnsureRunning(do *definitions.Do) error {
 	}
 
 	if !IsServiceRunning(srv.Service, srv.Operations) {
-		return ErrServiceNotRunning
+		e := fmt.Sprintf("The requested service is not running, start it with `eris services start %s`", do.Name)
+		return errors.New(e)
```
zramsay: http://play.golang.org/p/IQNzqzPXAY

commands/keys.go
compleatang: sounds good.

zramsay: on second thought, i'm gonna trash the option to export/import > 1 key at a time. it complicates dealing with filepaths and isn't really a primary use case...most users only need one key and, e.g., `keys gen` only makes a single key anyway. I'll make ADDR a required argument. bash scripts can do the heavy lifting...

compleatang:
Great!

PR #340, opened 2015-11-24
Chains new hot fix

compleatang:
cc @pietv can you give me a thumbsup on the changes to your merge function? 

The fundamental change is that for integers, strings, etc. the field in `over` should overwrite the field in `base`. 

This arose because, e.g., the `image` string in default.toml (base) was not getting properly overwritten by the `image` string in test_chain.toml (over).

Making the changes in 5f90fa0 solves that. All the tests seem to pass, but I just wanted you to be aware since you built that function.


pietv:
@compleatang yes, Merge didn't overwrite the values if they're were not empty and `Image` wouldn't as well.

This behaviour has been copied from from the `OverWrite` functions:

```
func OverWriteString(trumpEr, toOver string) string {
	if trumpEr != "" {
		return trumpEr
	}
	return toOver
}
```

compleatang:
Yeah. That was bad logic previously I'm pretty sure. The current behaviour is our desired behaviour for this function, no? (where `over` will overwrite `base` if not empty/zero) If that isn't desired behaviour then we can refactor around the Merge function to achieve the same result. Just depends how we want to do it.

pietv:
Totally agree.

compleatang:
Great. I'll merge this after those bigger things so nobody gets affected.

PR #348, opened 2015-11-26
Fetch DockerRunData logs

### LINKED ISSUE #343: Rewrite logs gathering for `data exec` and related tests
* Find a way to make it simpler:

```
		// now lets get the logs out
		// XXX: we only do this if the global config writer is a bytes.Buffer
		if config.GlobalConfig.Writer != nil {
			writer := config.GlobalConfig.Writer
			reader, ok := writer.(*bytes.Buffer)
			if !ok {
				return nil, nil
			}

			done := make(chan struct{}, 1)
			var b []byte
			go func() {
				// TODO: this routine will hang forever if  ReadAll doesn't complete
				// need to be smarter
				logger.Debugln("Attempting to read log reader.")
				b, err = ioutil.ReadAll(reader)
				done <- struct{}{}
			}()
			ticker := time.NewTicker(time.Second * 2)
		LOOP:
			for {
				select {
				case <-ticker.C:
					logger.Debugln("tick!")
					if reader.Len() == 0 {
						// nothing to read means dont bother waiting
						break LOOP
					} else {
						logger.Debugln("Read something", reader.Len())
					}
				case <-done:
					return b, err
				}
			}
		}
```

pietv:
Replace the `select` with the `waitContainer` call in `DockerRunData()` to make the function simpler.

Closes #343.

### DIFF
--- a/perform/docker_run.go
+++ b/perform/docker_run.go
@@ -5,13 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"os/signal"
 	"path/filepath"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/eris-ltd/eris-cli/config"
 	def "github.com/eris-ltd/eris-cli/definitions"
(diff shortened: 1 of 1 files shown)


compleatang:
:+1: Gonna wait for Zach to get keys finished so he doesn't have to rebase again, then will merge this.

pietv:
Of course! Thanks.

PR #355, opened 2015-11-27
Integration tests

ebuchman:
changes to eris-cli testing to
1) include integration tests (ie. tests defined in other repos and listed at https://github.com/eris-ltd/integration-tests/blob/master/test.sh#L5)
2) don't run all backends everytime

what's done here is the simplest approach, where the eris-cli tests run first (but only two of them, unless we're on `staging` or `master`), and then the integrations-test script launches a new box to run the integrations tests. It would be great to use one of the matdef boxes and save an extra couple of minutes, but not sure picking the box is trivial.


circle.yml
```diff
@@ -1,4 +1,10 @@
 machine:
+  environment:
+    GOPATH: /home/ubuntu/.go_workspace
+    REPO: ${GOPATH}/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+    INTEGRATION_TESTS_PATH: $HOME/integration-tests
+    INTEGRATION_TESTS_BRANCH: develop
+    BACKEND_TESTS_BRANCH: staging
```
compleatang: What's the difference between `BACKEND_TESTS_BRANCH` and `INTEGRATION_TESTS_BRANCH`?

circle.yml
```diff
@@ -1,4 +1,10 @@
 machine:
+  environment:
+    GOPATH: /home/ubuntu/.go_workspace
+    REPO: ${GOPATH}/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+    INTEGRATION_TESTS_PATH: $HOME/integration-tests
```
compleatang: what is this pointing to?

circle.yml
```diff
@@ -1,4 +1,10 @@
 machine:
+  environment:
+    GOPATH: /home/ubuntu/.go_workspace
```
compleatang: `$HOME/.go_workspace` will be less brittle

tests/test.sh
```diff
@@ -23,6 +23,7 @@ branch=${branch/-/_}
 #   installs by default on Linux"
 declare -a docker_versions18=( "1.8.0" "1.8.1" "1.8.2" "1.8.3" )
 declare -a docker_versions19=( "1.9.0" "1.9.1" )
+declare -a docker_latest_versions=("1.8.3" "1.9.1") # versions we run against no matter what
```
compleatang: This isn't going to work because there are diffs in APIs. See the current test output. Better to just grab the last element of the array if not doing a full run and run it through the current mechanism.

circle.yml
ebuchman: BACKEND we run all the backends, INTEGRATION we run the tests listed at the top of eris-ltd/integration-tests/test.sh

compleatang:
LGTM with the above notes / fixes.

compleatang:
@ebuchman let me know if you are going to work on those changes noted herein or if I should. I'd like to target this addition for rc2 if possible.

compleatang:
closing PR as it needed to be rebased and also included some additional modifications. using #378 instead.

PR #381, opened 2015-12-08
Deduplicate common flags

### LINKED ISSUE #271: deduplicate common flags
does it make sense to do this? seems like it would help with standardization (of helpers) & less technical debt.

zramsay:
- contracts was left alone because wonkyness with epm,
- flags that did not have a duplicate were not deduplicated into `flags.go` (but maybe they should, then all flags are in one place...?) and
- `func buildFlag()` in `flags.go` is ~messy and can probably be refactored to be clearer about which commands use which flags.

this PR closes #271


commands/old_falgs.go
```diff
@@ -0,0 +1,113 @@
+package commands
```
compleatang: is this file doing anything?

commands/contracts.go
```diff
@@ -78,7 +80,7 @@ deploy types.
 }
 
 //----------------------------------------------------
-
+// XXX todo deduplicate flags -> [zr] things get wonky with epm
```
compleatang: i can work through this tomorrow.

commands/old_falgs.go
zramsay: no. it's merely sentimental / easier to revert/debug when something breaks

commands/contracts.go
zramsay: most likely has to do with the wrapping

Issue #309, opened 2015-10-29
major init refactor


zramsay:
- [x] use toadserver for pulling in definition files
- [x] build test coverage #251 
- [x] `eris init --dir` -> #377, for testing (rejected)
- [x] `eris init --pull-images` -> #112, main images for tool pipeline
   - `eris/base`
   - `eris/keys`
   - `eris/data`
   - `eris/erisdb`
   - `eris/epm`
   - `eris/ipfs`
- [x] drop fresh priv_val/keypair, config.toml, & genesis.json -> #357  (relegated to `make`)
- [x] add a few "tour" directions (TODO, after `make`)

zramsay:
could not replicate but re-naming to more relevant issue

zramsay:
related to #251

compleatang:
related to #291

zramsay:
hm, for #357, we're gonna need some images (all save for `epm` and maybe `data`), so might as well pull 'em all! init should also compare latest quay images with local ones and pull in latest (see also #375)

compleatang:
My feeling is that it is best just to :dowit: once during that init (as long as we make sure that user's approve the giant bandwidth), so then it is done. It will take a while if they don't have anything but they can go `feed their marmots` or drink a cuppa... :cocktail: :coffee: :beers:

zramsay:
:yodawg: i hard you like pulling 5 gigs worth of images in `init`

Issue #377, opened 2015-12-08
`eris init --dir`


compleatang:
(mainly for testing purposes) we should be able to init into a directory other than ~/.eris.

zramsay:
closing in favour of task for #309

zramsay:
How important is this? Because the init sequence is tied to /common and thus `ResolveErisRoot()`, it all has to be unbundled (a bunch of ~unnecessary code). If this is only for tests, we already have `config.ChangeErisDir(erisDir)` with `var erisDir = path.Join(os.TempDir(), "eris")` as part of the `TestsInit` sequence in `testutils` which leaves your `~/.eris` alone

compleatang:
This is low priority. We may want to refactor `ResolveErisRoot()`, but we dont' have to do it today. 

The problem is that during the testing sequence we first `eris init --yes` so that we can start IPFS which is needed to be running for the first tranche of package tests (I think). If we reworked, predominantly `files_tests.go` so that it did not expect IPFS running then we could remove that `eris init --yes` in test_tool.sh which would stop messing with my local ~/.eris directory (which is what was happening when I filed the issue).

zramsay:
makes sense

Issue #357, opened 2015-11-28
eris init should drop a priv_validator and keypair into the proper location


compleatang:
folks should have a unique genesis.json in their `chains/default` directory with their shiny new key rather than current ultra hacky system.

zramsay:
closing in favour of task for #309

zramsay:
will be moved to `make` because :imp: import cycles

Issue #251, opened 2015-09-18
Build test coverage for `initialize` package


zramsay:
requires #309 / running toadserver

compleatang:
related: #83 #257

zramsay:
closing in favour of task for #309

PR #396, opened 2015-12-14
Init refactor

zramsay:
This refactor has init using the toadserver (http://ipfs.erisbootstrap.sexy:11113) to pull in all required service/action/chain definition files. These files are tracked at (http://pinkpenguin.interblock.io:46657/list_names). For redudancy (if toadserver is down), the `--source=rawgit` flag can be set to get files from GitHub instead. In fact, the tests (#251) compare all the files from each source; any updates need to happen on both the chain and GH. 

Default docker images required for the eris tool are also pulled in, with clear warning (#112) .

closes: #309 
note: #377 is rejected (as is #357)


initialize/default_services.go
```diff
@@ -1,65 +0,0 @@
...
-
-func DefaultKeys() string {
-	return fmt.Sprintf(`[service]
-name = "keys"
-image = "quay.io/eris/keys"
-data_container = true
-exec_host = "ERIS_KEYS_HOST"
-`)
-}
-
```
zramsay: will move these to tests - to test against both 1) files got from toadserver and 2) files cloned from each `eris-services/actions`. This will add redundancy to keeping the definition files in check

pietv:
Awesome :+1:

compleatang:
:+1:

PR #407, opened 2015-12-28
Logrus logger

pietv:
The Logrus library replaces the current logging library to allow to get rid of formatting in log messages (tabs, newlines, `=>`) and to make the code look cleaner.

## Library usage

Simple logging at various levels:

```
log.Fatalf("Error changing directory: %v")
log.Error(err)
log.Warn("Failed to get that file. Sorry")
log.Info("Chain container does not exist")
log.Debug("Not requiring input. Proceeding")
``` 

Logging with a stand out item (unnamed). That item is most often a container, a chain, or a service name.

```
log.WithField("=>", doChns.Name).Debug("Starting chain")
``` 

Logging with a named stand out parameter. 

```
log.WithField("args", do.Operations.Args).Info("Performing action (from tests)")
log.WithField("image", srv.Image).Debug("Container does not exist. Creating")
log.WithField("drop", dropped).Debug()
```

Logging with multiple standout parameters.

```
log.WithFields(log.Fields{
	"from": do.Name,
	"to":   do.NewName,
}).Info("Renaming action")
  
log.WithFields(log.Fields{
  	"=>":        servName,
	"existing#": containerExist,
	"running#":  containerRun,
}).Info("Checking number of containers for")
```
  
Standout item names are sorted alphabetically with the unnamed parameter `=>` always the first. The order can be changed by splitting one log message into multiple log messages:
  
```
log.WithField("running#", containerRun).Info()
log.WithField("existing#", containerExist).Info()
log.WithField("=>", servName).Info("Checking number of containers for")
```
  
### Eris' default log levels

Default Eris log level for command line tools has been set to `warn`.

```
log.SetLevel(log.WarnLevel)
if do.Verbose {
	log.SetLevel(log.InfoLevel)
} else if do.Debug {
	log.SetLevel(log.DebugLevel)
}
```

### Test prologue

```
log.SetFormatter(logger.ErisFormatter{})

log.SetLevel(log.ErrorLevel)
// log.SetLevel(log.InfoLevel)
// log.SetLevel(log.DebugLevel)
```

### Writer

Logrus library doesn't give access to its writer, so the writer has been set to `stdout`:

```
if log.GetLevel() > 0 {
  err = ipfs.GetFromIPFS(s[1], fileName, "", os.Stdout)
} else {
  err = ipfs.GetFromIPFS(s[1], fileName, "", bytes.NewBuffer([]byte{}))
}
```

### A few things for consideration

* Logrus' `Print` level equals `Info` level in the new library and is printed out only when the `verbose` or `debug` mode is on, so older `log.Print()` log messages have been converted to `log.Warn()` (or to `log.Error()` where it made more sense) with a few exceptions.

* Logrus always ends a log message with an `\n`, so there's no need to invoke `log.Infoln()` or `log.Infof("\n")` or similar. For this reason, the messages that require user's input (**Do you wish to continue? (y/n):**) have been converted to use `fmt.Print()`.

* Logrus logs only to `stderr`.

* Logrus's comma operator doesn't add a space between operands (a small bug on the library's part). This snippet:

  ```
  log.Print("a", "b")
  ```
  
  produces `ab\n`. (While it should have produced `a b\n`)

## Log messages

These rules have been applied to log messages while rewriting them in some places:

* Sentences start with a capital letter, but don't end with a dot. Multiple sentences on the same line are separated with a dot: **Chain not currently running. Skipping** 
* Articles in log messages are usually dropped (magazine headline style); separate words are not capitalized: **Getting connection details from environment**
* Names of functions are removed, but the messages are made still easily distinguishable (greppable): **Running data container** (instead of **DockerRunData**) or **Table report initialized** (instead of **PrintTableReport Initialized**).
* Verbs for log messages; nouns for tags (the `WithField` thing). The noun log messages are converted to verbs or tags:

  ```
  log.WithField("drop", dropped).Debug()
  ```
  
  instead of

  ```
  logger.Debugf("Args to add to the steps =>\t%s\n", dropped)
  ```
  
* Removed all initial indents: **"Blah-blah-blah"** instead of **"\tBlah-blah-blah"**.
* All references to container `id`s have been removed or replaced with container full names.
* Log messages in tests have been given the suffix **from tests** (vs **via tests**) for more uniformity.

## Color scheme

256color

The yellow on the screenshot is not a color, but the bold terminal attribute which allows for more flexibility with various user's color schemes and terminal settings. 

Errors and warnings are not using special colors (tried a few—abandoned; they make the output look distracting and cheaper).

Reversed fore- and background and a different Terminal color setting for bold:

reversed

This is Linux console on Digital Ocean (its stand out color is white):

linux

Dark grey is the only color with this color scheme. It is neutral enough and can match many user defined color schemes. Some modern terminals adjust the colors to prevent them from becoming invisible (Mac OS Terminal app):

Darker background:

darkgrey1

Lighter background:

darkgrey2

On terminals with 8 colors:

```
$ TERM=linux eris services start ipfs -d
```
8colors

On terminals with no color support or Windows:

```
$ TERM=dumb eris services start ipfs -d
```
0colors
 
## Other changes

* The tests for `ImportService`, `ExportService`, `PutFiles`, `GetFiles` have been rewritten to use a fake service implementation which records which API endpoints have been called and responds with a fixed response. The tests check that the functions call the server and that they do what they're supposed to do with received results (return a hash, create a file, etc.) This way it is much more faster and reliable and still tests both Eris CLI and `common/go/ipfs` library. Most importantly, it will allow to get rid of IPFS startup errors once and for all. The only downside is that if the IPFS API interface changes, the Eris CLI tests wouldn't know that.
* The `TestServiceLinkChainedService()` test has been amended for more robustness.
* The `perform` package functions have been changed to use container full names instead of IDs.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -117,6 +117,10 @@
 		{
 			"ImportPath": "gopkg.in/yaml.v2",
 			"Rev": "5d6f7e02b7cdad63b06ab3877915532cd30073b4"
+		},
+		{
+			"ImportPath": "github.com/Sirupsen/logrus",
+			"Rev": "ad63b06ab3877915532cd30073b45d6f7e02b7cd"
 		}
 	]
 }
--- a/Godeps/_workspace/src/github.com/Sirupsen/logrus/.travis.yml
+++ b/Godeps/_workspace/src/github.com/Sirupsen/logrus/.travis.yml
@@ -1,0 +1,7 @@
+language: go
+go:
+  - 1.3
+  - 1.4
+  - tip
+install:
+  - go get -t ./...
--- a/Godeps/_workspace/src/github.com/Sirupsen/logrus/CHANGELOG.md
+++ b/Godeps/_workspace/src/github.com/Sirupsen/logrus/CHANGELOG.md
@@ -1,0 +1,55 @@
+# 0.9.0 (Unreleased)
+
+* logrus/text_formatter: don't emit empty msg
+* logrus/hooks/airbrake: move out of main repository
+* logrus/hooks/sentry: move out of main repository
+* logrus/hooks/papertrail: move out of main repository
+* logrus/hooks/bugsnag: move out of main repository
+
+# 0.8.7
+
+* logrus/core: fix possible race (#216)
+* logrus/doc: small typo fixes and doc improvements
+
+
+# 0.8.6
+
+* hooks/raven: allow passing an initialized client
+
+# 0.8.5
+
+* logrus/core: revert #208
+
+# 0.8.4
+
+* formatter/text: fix data race (#218)
+
+# 0.8.3
+
+* logrus/core: fix entry log level (#208)
+* logrus/core: improve performance of text formatter by 40%
+* logrus/core: expose `LevelHooks` type
+* logrus/core: add support for DragonflyBSD and NetBSD
+* formatter/text: print structs more verbosely
+
+# 0.8.2
+
+* logrus: fix more Fatal family functions
+
+# 0.8.1
+
+* logrus: fix not exiting on `Fatalf` and `Fatalln`
+
+# 0.8.0
+
+* logrus: defaults to stderr instead of stdout
+* hooks/sentry: add special field for `*http.Request`
+* formatter/text: ignore Windows for colors
+
+# 0.7.3
+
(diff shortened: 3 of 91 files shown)


compleatang:
So epic!

PR #412, opened 2016-01-08
data directory & command (minor refactor)

### LINKED ISSUE #400: Move away from using `~/.eris/data`
**Problem** -- `~/.eris/data` is a potential source of confusion to users.

**Proposed Fix 1** -- the scratch directory is pretty self-explanatory and already is used by other eris tools as a centralized cache point. eris-cli follow this paradigm, which would have the added benefit of removing a possible source of confusion. Proposal 1 would change the ~/.eris/data to ~/.eris/scratch/data-containers (or something along those lines, am not married to `data-containers`). This fix (versus Fix 2) has the benefit of being much more simple and would keep to the current paradigm.

**Proposed Fix 2** -- do away with a caching directory at all. This would be a much larger change as we will need to make sure we *require* a source flag to be sent with any data operation. There is also some funkiness that happens when we *dump* an ~/.eris directory to the host machine because there is a bunch of latent folders which get pulled in. We currently use a temp directory to handle most of this, but there is still some challenges here because we could end up dumping a whole bunch of empty directories into the destination on data export.

Between the two, my vote is to utilize `Fix 1` in the short term and perhaps move to `Fix 2` later.

zramsay:
this PR:
- eliminates use of `~/.eris/data` (replaced by `~/.eris/scratch/data` under the hood)
- refactors `eris data import/export` to each require `src/dest`
- 3 argument are required: `eris data import/export NAME SRC DEST`
- helpers updated to reflect this change, with notes on path specification.
- data package tests are updated to be independent of each other.

closes #400

### DIFF
--- a/commands/data.go
+++ b/commands/data.go
@@ -2,7 +2,7 @@ package commands
 
 import (
 	"fmt"
-	"path/filepath"
+	//	"path/filepath"
 	"strings"
 
 	"github.com/eris-ltd/eris-cli/data"
@@ -15,6 +15,7 @@ import (
 //----------------------------------------------------
 
 // Primary Data Sub-Command
+//TODO re-write
 var Data = &cobra.Command{
 	Use:   "data",
 	Short: "Manage data containers for your application.",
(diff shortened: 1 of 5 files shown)


compleatang:
LGTM.

compleatang:
Merging those other branches requires this to be rebased. Sorry about that. Thought I merged them in the right order. Also can you switch the swarms back?

Issue #393, opened 2015-12-14
Chains new should not fail if data container exists. 


compleatang:
**Problem** -- if a user has started and stopped a chain and then has edited their genesis.json or some other file and then does `eris chains new testChain` to pull those files into the data container for the chain, the command will fail. Since `chains new` is the shortcut way to get files into the data container and then start the chain this is an undesirable result from a ux perspective. 

**Current WorkAround** -- `eris data rm testChain && eris chains new testChain`

**Proposed Fix** -- Although it is easy enough to `eris data rm testChain && eris chains new testChain`, it would remove a potential pain point for users if they could just `eris chains new testChain` and have it not fail during the data container creation portion of the sequence.

PR #413, opened 2016-01-08
make `eris chains new` work when data container already exists

pietv:
This PR fixes the issue #393 by checking container existence prior to creating it.

The `data.IsKnown()` call is replaced with `util.IsDataContainer()`.

The `data.IsKnown()` function and its helpers are removed.


compleatang:
LGTM :+1:

Issue #291, opened 2015-10-17
`update --tool` should be able to fetch the eris-cli binary if go not available


compleatang:

zramsay:
`init` could use this as well. should probably prompt user too.

compleatang:
blocked until #309 is finished.

compleatang:
thanks to @mxjxn 's work we can close :)

PR #438, opened 2016-01-15
Issue #291

mxjxn:
https://github.com/eris-ltd/eris-cli/issues/291

Modified CheckGitAndGo() to return two boolean values for whether git and go are found on the system.

Modified UpdateEris() to take two extra parameters specifying whether to check git and go respectively.

Added downloadLatestBinary() which downloads the latest release of eris-cli to the users ~/bin folder, and tells them to include that folder in their PATH variable. (Perhaps it should check for it and, if missing, ask the user if they'd like to add it?)

Also added tests for CheckGitAndGo and downloadLatestBinary functions.

### DIFF
--- a/commands/update.go
+++ b/commands/update.go
@@ -26,6 +26,6 @@ func addUpdateFlags() {
 }
 
 func UpdateTool(cmd *cobra.Command, args []string) {
-	util.UpdateEris(do.Branch)
+	util.UpdateEris(do.Branch, true, true)
 
 }
--- a/util/update_tool.go
+++ b/util/update_tool.go
@@ -2,19 +2,109 @@ package util
 
 import (
 	"fmt"
+	log "github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/Sirupsen/logrus"
+	"github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/eris-ltd/common/go/common"
+	"io"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
-
-	log "github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/Sirupsen/logrus"
-
-	"github.com/eris-ltd/eris-cli/Godeps/_workspace/src/github.com/eris-ltd/common/go/common"
+	"runtime"
+	"strings"
 )
 
-func UpdateEris(branch string) {
+func downloadLatestRelease() (string, error) {
+	latestURL := "https://github.com/eris-ltd/eris-cli/releases/latest"
+	resp, err := http.Get(latestURL)
+	if err != nil {
+		log.Printf("could not retrieve latest eris release at %s", latestURL)
+	}
+	latestURL = resp.Request.URL.String()
+	lastPos := strings.LastIndex(latestURL, "/")
+	version := latestURL[lastPos+1:]
+	platform := runtime.GOOS
+	arch := runtime.GOARCH
+	hostURL := "https://github.com/eris-ltd/eris-cli/releases/download/" + version + "/"
+	filename := "eris_" + version[1:] + "_" + platform + "_" + arch
+	fileURL := hostURL + filename
+	switch platform {
+	case "linux":
+		filename += ".tar.gz"
+	default:
+		filename += ".zip"
+	}
+
+	ChangeDirectory("bin")
+	var erisBin string
+	output, err := os.Create(filename)
+	// if we dont have permissions to create a file where eris cli exists, attempt to create file within HOME folder
+	if err != nil {
+		erisBin := filepath.Join(common.ScratchPath, "bin")
+		if _, err = os.Stat(erisBin); os.IsNotExist(err) {
+			err = os.MkdirAll(erisBin, 0755)
+			if err != nil {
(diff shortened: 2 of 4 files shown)


util/update_tool.go
```diff
@@ -2,19 +2,82 @@ package util
...
+	defer fileResponse.Body.Close()
+
+	// save file-contents to new file
+	io.Copy(output, fileResponse.Body)
+	if err != nil {
+		log.Println("Error while downloading", filename, "-", err)
+		return "", err
+	}
+
+	log.Println("downloaded eris binary", version, "for", platform, "to ~/bin directory. Please ensure ~/bin is in your PATH variable.")
```
compleatang: rather than being opinionated about ~/bin let's use [exec.LookPath](https://golang.org/pkg/os/exec/#LookPath) to find where eris is installed.

commands/update.go
```diff
@@ -26,6 +26,7 @@ func addUpdateFlags() {
 }
 
 func UpdateTool(cmd *cobra.Command, args []string) {
-	util.UpdateEris(do.Branch)
+	var checkGit, checkGo bool = true, true
```
compleatang: unless we're gonna reuse these variables (perhaps as flags, although I'm not sure the utility of that) just give the function `(do.Branch, true, true)`

util/update_tool.go
```diff
@@ -2,19 +2,82 @@ package util
...
+	resp, err := http.Get(latestURL)
+	if err != nil {
+		log.Printf("could not retrieve latest eris release at %s", latestURL)
+	}
+	latestURL = resp.Request.URL.String()
+	lastPos := strings.LastIndex(latestURL, "/")
+	version := latestURL[lastPos+1:]
+	platform := runtime.GOOS
+	hostURL := "https://github.com/eris-ltd/eris-cli/releases/download/" + version + "/"
+	filename := "eris_" + version[1:] + "_" + platform + "_amd64"
```
compleatang: let's use runtime.GOARCH instead of `_amd64` ...

util/update_tool.go
```diff
@@ -2,19 +2,82 @@ package util
...
+		log.Println("Go is not installed. Downloading eris-cli binary...")
+		_, err := downloadLatestRelease()
+		if err != nil {
+			log.Println("Latest binary failed to download with error:", err)
+			os.Exit(1)
+		}
+
+	}
+	if hasGit == false {
+		os.Exit(1)
```
compleatang: what's your reasoning here @mxjxn ...? whether logic is good or bad, still need to tell the user why we're exiting. in general if we're gonna `os.Exit(1)` on users we need to give the user the reason we're exiting on them, as well as a fix (if there is an easy answer, here it would be... install go (link) )....

util/update_tool.go
```diff
@@ -39,35 +102,60 @@ func UpdateEris(branch string) {
...
 
-func ChangeDirectory() {
-	goPath := os.Getenv("GOPATH")
-	if goPath == "" {
-		log.Fatal("You do not have $GOPATH set. Please make sure this is set and rerun the command")
-	}
+func ChangeDirectory(to string) {
+	if to == "bin" {
+		usr, err := user.Current()
+		userBin := usr.HomeDir + "/bin"
```
compleatang: this is quite opinionated and for non-OSX users non-standard. many OSX users do this of course. If we use `exec.LookPath` then we should be able to install to the directory where `eris` is already installed rather than having to perform this check...

util/update_tool.go
```diff
@@ -39,35 +102,60 @@ func UpdateEris(branch string) {
...
+			mkErr := os.Mkdir(userBin, 0755)
+			if mkErr != nil {
+				log.Fatalf("error creating directory: %v", mkErr)
+			}
+		}
+		err = os.Chdir(userBin)
+		if err != nil {
+			log.Fatalf("Error changing directory: %v", err)
+		}
+		log.WithField("dir", "~/bin").Debug("Directory changed to")
```
compleatang: `log.WithField("dir", userBin).Debug("Directory changed to")`

util/update_tool.go
mxjxn: I didnt have a good reason but it makes sense that I should be instructing the user to install Go, if they dont have it.

util/update_tool.go
mxjxn: I was uncertain as to how to handle this. As far as i understand, the binary (if go isnt installed) would be installed in /usr/bin, but the user wouldnt have access from the eris-cli tool. 

I wasnt aware of exec.LookPath but that would make sense to find an already-installed binary.

compleatang: No worries. Had to look it up myself :)

util/update_tool.go
compleatang: No worries. I mean `go` requires `git` so it would be rare that this would not be triggered... To ensure that you can change the `if` to an `else if`. 

Adding a simple message there would be helpful tho.

compleatang:
LGTM. @mxjxn one last thing, could you rebase and squash these commits so they look nicer in the commit history please?

mxjxn:
sorry about the delay. had to give myself a crash course in rebase.

Issue #376, opened 2015-12-08
refactor listing functions


compleatang:
- [zr]
- [ ] replace the `Type` field in the listing functions with `Machine` using [libmachine](https://godoc.org/github.com/docker/machine/libmachine/host) #371 (defered)
- [x] `eris keys ls` => #401 
- [x] `eris data ls` => for host/container (#339) [zr]: rejected because #400

zramsay:
oh maybe we replace that header with `MACHINE` for #371

compleatang:
nice. like it.

zramsay:
closed by #457l; doesn't include machine field...postponed until `remotes ls` takes hold.

PR #457, opened 2016-01-26
ls refactor

zramsay:
mix of prettier output for user & cleaner code under the hood. 
- primary listing funcs moved from `util` to new pkg `list` (more could go, IMO)
- undid some excessive deduplication in favour of clarify
- deduplicated `do.Quiet` & `do.Operations.Args = []string{"testing"}`
- removed `CONTAINER #` field from table; redundant with container name (`_1`)

addresses #376 save for the `MACHINE` field, which is left as a placeholder. That's gonna block until `eris remotes ls` alongside the libmachine's persist/filestore is grounded.

requires still:
- [x] test for `data ls` (& maybe the `list` pkg itself)
- [x] finish bullet point 1.


commands/data.go
```diff
@@ -130,29 +130,18 @@ var dataRm = &cobra.Command{
...
 //----------------------------------------------------
 func ListKnownData(cmd *cobra.Command, args []string) {
-	do.Existing = true
-	if err := util.ListAll(do, "data"); err != nil {
+	if err := list.ListDatas(do); err != nil {
 		return
 	}
-
-	// https://www.reddit.com/r/television/comments/2755ow/hbos_silicon_valley_tells_the_most_elaborate/
-	datasToManipulate := do.Result
```
compleatang: RIP `datasToManipulate`

compleatang:
nice and clean. this LGTM. will do the merges tomorrow but :+1:

commands/data.go
zramsay: ahaha I hesitated for awhile on this one

PR #465, opened 2016-01-28
Fixing the directory calculation on Windows

pietv:
This PR implements a fix for the 'Error changing directory: chdir c:\bin\eris`.


compleatang:
:+1: LGTM. totally missed that code before.

PR #466, opened 2016-01-28
deduplicate common files & images

### LINKED ISSUE #458: images should be variables located in one place only
images we use are littered throughout the code. they should be in one place.

ERISDB_IMAGE = ....
ERISPM_IMAGE = ...
ERISCM_IMAGE = ...

etc.

zramsay:
- moves image names throughout the code to `version/images.go`
- moves service/action definition names to `version/files.go` ++ cleaner tests
- failover to dockerhub if err pulling from quay (but only on init)

closes #458


compleatang:
:+1:

PR #467, opened 2016-01-29
spf13 lib updates; cast, cobra, pflag, viper

zramsay:
most obvious change noticed: uglier flag helper ... default moved to end of flag definition (in brackets)


zramsay:
Note: this update to the `viper` lib required manual modification: packages `github.com/hashicorp/hcl` and `gopkg.in/fsnotify.v1` were not vendored by `viper`. They don't appear don't appear needed for our purposes. These packages and any calls to them were commented out without issue (i.e., tests pass).

PR #476, opened 2016-02-01
lots of little organizational cleanup

compleatang:
changes:

* move "projects" to "apps" to simplify onboarding and cohesive
* moved files that used to be in `commands` to `cmd` in keeping with the rest of our repo's organizational standards (also that directory was superfluous in reality).
* moved files from `testutils` to `tests` for cohesiveness (currently only bash scripts and fixtures are in tests so no conflicts with go packages).
* removed the `TESTS_IN_CIRCLE` env var calls since we use remote backends now those are no longer needed and just add mental overhead for folks getting up to speed on using.
* mocked out agent which will be needed very soon (marketplace application will need)


compleatang:
because this PR cuts across the packages can @pietv  and @zramsay both give me a :+1: that its OK. don't want to merge this too quickly and it isn't super essential anyway, mostly just housekeeping.

pietv:
@compleatang please can you hold back the merge for one day, because i'm planning to merge the refactor for chains and services tests today, and in case if something fails I might be anxious to understand why it failed (on osx and win), especially due to the TEST_IN_CIRCLE variable removal.
I'm okay to rebase and tests this PR afterwards.

:+1:

compleatang:
Yup. Not in a hurry and that's why I wanted to check.

zramsay:
LTGM. lovin how much cleaner this is

PR #500, opened 2016-02-06
exec functions signatures

### LINKED ISSUE #449: Change the signature of Exec functions to return container logs upon exit
This is a common usage pattern to use container output in other functions. Currently it is done by initializing the `GlobalConfig` with a buffer and then use that buffer as container output.

This is a proposal to change the functions signatures to return container output:
* `func DockerExecData(...) (out []byte, err error)`
* `func DockerExecService(...) (out []byte, err error)`

As an alternative `out` could be a string.

What do you guys @zramsay, @compleatang think?

pietv:
This PR changes the signatures of Exec functions to return logging output instead of having to set the global config yourself. This closes #449.

* DockerExecData(ops *def.Operation, service *def.Service) (buf *bytes.Buffer, err error)
* DockerExecService(srv *def.Service, ops *def.Operation) (buf *bytes.Buffer, err error)
* ExecData(do *definitions.Do) (buf *bytes.Buffer, err error)
* ExecService(do *definitions.Do) (buf *bytes.Buffer, err error)
* ExecChain(do *definitions.Do) (buf *bytes.Buffer, err error)

`bytes.Buffer` return type is used for versatility: it can be used as a string (`buf.String()`), a slice of bytes (`buf.Bytes()`), or a reader/writer (simply `buf`). There are places in the code that require all of these.

Two global writers have been added to the `config` package:

```
config.GlobalConfig.InteractiveWriter     
config.GlobalConfig.InteractiveErrorWriter
```

They are introduced to allow working with exec containers interactively (for example,
when running a `bash` shell) in addition to returned container log which is returned at the end of Exec function. In all other cases they're shut.

* `perform/docker_run.go` > `perform/perform.go`.

* `io.Copy(config.GlobalConfig.Writer, buf)` has been added to `keys` and `chains make` package fuctions, because they use global buffer externally (mostly via tests).


services/services_test.go
```diff
@@ -163,20 +163,18 @@ func TestExecService(t *testing.T) {
...
 
-	if err := ExecService(do); err != nil {
+	buf, err := ExecService(do)
+	if err != nil {
 		t.Fatalf("expected to execute service, got %v", err)
 	}
 
-	if out := buf.String(); !strings.Contains(out, ".bashrc") {
-		t.Fatalf("expected a file in the exec output, got %v", out)
+	if !strings.Contains(buf.String(), ".bashrc") {
```
compleatang: much cleaner this way.

compleatang:
As always, great work. LGTM. This will be quite helpful for testing as well as agents purposes. Also its a hell of a lot cleaner than the writer trickery we were doing before.

PR #528, opened 2016-02-21
Remote logging to Bugsnag

### LINKED ISSUE #403: error dumping
docker-machine recently did this...:

```
Error Reporting

As part of our continued effort to make the Docker Machine experience better, we have added some behavior which will report various kinds of failures to Bugsnag for the core maintainers to debug.

In most cases this should happen in the background without needing any input or action from the user. If you wish to opt out of this error reporting, you can create a no-error-report file in $HOME/.docker/machine, e.g.:

$ mkdir -p ~/.docker/machine && touch ~/.docker/machine/no-error-report
```

That may be a smart idea for us. Potentially, when a user encounters and error it would automatically send a notification to bugsnag, with the full debug output (assuming we'd flush the debug level logs, perhaps to a different writer...?). 

This may be connected to the move to logurus.

For reference:
https://bugsnag.com/docs/notifiers/go

### LINKED ISSUE #506: Turn formatting off for redirected logging messages
Redirected output (for example, `eris init > init.txt` as in issue #505) clutters the debugging output.

### LINKED ISSUE #523: Ungodepped dependencies on a clean `eris-cli` install
On a system with a clean GOPATH directory, installing the Eris CLI with the `go get github.com/eris-ltd/eris-cli/cmd/eris` command reveals that some of the packages not point properly to the Godeps locations:

```
$ ls -l /cygdrive/c/src/github.com
total 0
drwxrwx---+ 1 Peter Peter 0 Feb 17 17:03 Azure
drwxrwx---+ 1 Peter Peter 0 Feb 17 17:03 docker
drwxrwx---+ 1 Peter Peter 0 Feb 17 17:03 eris-ltd
drwxrwx---+ 1 Peter Peter 0 Feb 17 17:03 Sirupsen
```

pietv:
### Bugsnag

This PR implements a remote logger to be used to send `panic()` events along with debug logging output for further analysis.

Bugsnag is selected as a remote logger because:  1) Unlike Papertrail it has more convenient and stable Go client library  and allows for more structured view of the sent events. 2) Unlike Sentry it is less expensive and, again, allows for more structured view of information. 

View of a panic stack trace:

crash report

`eris -d` output (irrespective of the console logging level, remote logging is always at the debug level):

debug output

Additional bits of information in separate tabs and on separate rows:

docker client

This PR also adds the `CrashReport` variable to the `eris.toml` definition file which defaults to "bugsnag" to control the remote logger behavior. Setting it to anything else but "bugsnag" will turn the remote logging off.

Notifying the users in `eris init` to allow them to opt out from remote logging like this:

```
Eris sends crash reports to a remote server in case something goes completely 
wrong. You may disable this feature by adding the CrashReport = "don't send" 
line to the /Users/peter/.eris/eris.toml definition file.
```
This is opinionated: can be changed to opt in easily, but assuming users are lazy and we don't send anything yet but panics, this is not a huge privacy intervention.

Currently the logging library uses an `APIKey` authentication key of a personal Bugsnag account. If Bugsnag is okay for remote logging, we'd need to create a team/team account there (or share a password). Before merging a temporary token can be set like this for testing: 

```
$ export ERIS_BUGSNAG_TOKEN=...
```

Closes #403.

### Other fixes

* Setting default logging output stream to stdout instead of stderr for more convenient redirecting and viewing of logging information.

* Turning off the color formatting when redirecting output to a file `eris > out` or a viewer `eris|more`. Fixes #506.

* Cleaning up ungodepped dependencies on Windows. Fixes #523.

* small editing of the `eris init` screen to fit the 80-column console width.

* small fixes satisfying `go vet ./...`.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -133,6 +133,31 @@
 		{
 			"ImportPath": "github.com/docker/go-units",
 			"Rev": "f49f7ea428ed3f42a846917e4a4856fcd165f6bc"
+		},
+		{
+			"ImportPath": "github.com/Sirupsen/logrus",
+			"Comment": "v0.8.7-53-g446d1c1",
+			"Rev": "446d1c146faa8ed3f4218f056fcd165f6bcfda81"
+		},
+		{
+			"ImportPath": "github.com/docker/docker/pkg/system",
+			"Rev": "f7e02b752bcf687aab0120e6b06ab38779371ec0"
+		},
+		{
+			"ImportPath": "github.com/Azure/go-ansiterm",
+			"Rev": "99cbe6fbe8d074bcf687aab012c3c0d5297f614a"
+		},
+		{
+			"ImportPath": "github.com/bugsnag/bugsnag-go",
+			"Rev": "ab8487ef058074bc60171a6436fabcfac287a001"
+		},
+		{
+			"ImportPath": "github.com/bugsnag/panicwrap",
+			"Rev": "40098faa760171a6436fab017853031334b3c8b9"
+		},
+		{
+			"ImportPath": "github.com/bugsnag/osext",
+			"Rev": "cd2b779537780171a6436fbc39950e081b0187bb"
 		}
 	]
 }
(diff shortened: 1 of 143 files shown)


compleatang:
LGTM.. 

Bugsnag will work well for these purposes I think and we have an eris account now. Great work as always!

compleatang:
just needs a quick rebase.

PR #543, opened 2016-03-01
proposed compilers url strategy

compleatang:
**Problem** -- currently we have no way to version the information which comes and goes to the compilers service. which means we have no way to migrate changes to how epm talks to the compilers. 

normally this is fine cause we run nightly builds of the image which pulls in the latest from eth's ppas and leaves our go wrapper code fine. that go wrapper code is due for a refresh ( https://github.com/eris-ltd/eris-compilers/issues/51 ). in turn this will affect how epm talks to the compilers. 

as contract bundles and engines get more complex we will need to ensure that the epm and compilers interface is somewhat coupled. 

**Solution** -- better version the connection between epm and compilers wrapper

**Strategy** -- for discussion

(option a) Use a ports strategy as outlined here. Just run a bunch of compilers in parallel.
(option b) (better option if we have manpower) Improve the handshake between eris-pm and the compilers.
(option c) ..?

cc @eris-ltd/team-platform

### DIFF
--- a/cmd/pkgs.go
+++ b/cmd/pkgs.go
@@ -1,9 +1,13 @@
 package commands
 
 import (
+	"fmt"
 	"os"
+	"strconv"
+	"strings"
 
 	"github.com/eris-ltd/eris-cli/pkgs"
+	"github.com/eris-ltd/eris-cli/version"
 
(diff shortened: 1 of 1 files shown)


zramsay:
NOTE: app-05 (eris-pm) fails on travis only

PR #555, opened 2016-03-06
updates to init; the checker was returning on first query and better …

compleatang:
…display of errors


zramsay:
fixture failures, merging

PR #557, opened 2016-03-06
chains new; overwrite scratch dir by default

### LINKED ISSUE #540: eris chains new should *not* respect the data dir
lots of times stuff gets "stuck" in that directory and even advanced users (i.e., me) forget to check that.

new should mean new and we should forcibly overwrite the scratch dir (or, better, not use and just use the new data dest|src paradigm).

zramsay:
- prevents bad things from happening when new-ing (then removing) chains of the same name
- closes #540

### DIFF
--- a/chains/chains_test.go
+++ b/chains/chains_test.go
@@ -285,7 +285,9 @@ func TestCatChainContainerGenesis(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


zramsay:
fixtures failure (eris-pm): ignore for now

PR #566, opened 2016-03-08
import priv_validator to eris-keys; closes #459

### LINKED ISSUE #459: chains new improvements
- [x] make sure priv_validator is imported to keys
- [x] remove the cruft of the old sprawling flags 

new should instantiate the chain locally based on files rather than have responsibilities for `make`ing any of the necessary files. control the new process a bit closer at the cli level so that when lower level containers start we do not need to rely on them to give us good error messages.

compleatang:


compleatang:
@pietv will leave this to you to merge.

Issue #295, opened 2015-10-19
`eris clean --uninstall`


zramsay:
nice-to-haves that aren't priority. bumping to backlog. from #209

compleatang:
closing in favor of #479

zramsay:
re-opening as backlog. not really priority imo

PR #573, opened 2016-03-11
refactor `eris clean` sequence

### LINKED ISSUE #479: some love for `clean`
mostly the user friendliness & better prompting & tests

zramsay:
- closes #479 but not #295 
- prompts user appropriately
- cleaner test + some other cleanings
- catches dangling data containers:w

### DIFF
--- a/clean/clean.go
+++ b/clean/clean.go
@@ -6,7 +6,16 @@ import (
 )
 
 func Clean(do *definitions.Do) error {
-	if err := util.Clean(do.Yes, do.All, do.RmD, do.Images); err != nil {
+	// in util so that other pkgs can import it easily
+	toClean := map[string]bool{
+		"yes":        do.Yes,
+		"all":        do.All,
+		"containers": do.Containers,
+		"scratch":    do.Scratch,
+		"rmd":        do.RmD,
+		"images":     do.Images,
+	}
+	if err := util.Clean(toClean); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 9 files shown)


clean/clean_test.go
```diff
@@ -3,61 +3,90 @@ package clean
...
+	// start a bunch of eris containers
+	testStartService("ipfs", t)
+	testStartService("tor", t)
+	testStartService("keys", t)
+
+	testStartChain("dirty-chain0", t)
+	testStartChain("dirty-chain1", t)
+
+	testCreateDataContainer("filthy-data0", t)
+	testCreateDataContainer("filthy-data1", t)
```
compleatang: `filthy`

clean/clean.go
```diff
@@ -6,7 +6,16 @@ import (
 )
 
 func Clean(do *definitions.Do) error {
-	if err := util.Clean(do.Yes, do.All, do.RmD, do.Images); err != nil {
+	// in util so that other pkgs can import it easily
+	toClean := map[string]bool{
+		"yes":        do.Yes,
+		"all":        do.All,
+		"containers": do.Containers,
+		"scratch":    do.Scratch,
```
compleatang: Since scratch and containers are default true, recommend they render here as:

```go
"contaienrs": !do.Containers,
"scratch":      !do.Scratch,
```

That way the util.Clean function won't have to account for the defaults set in the cmd package level (or better yet do that transform in cmd/clean.go .

compleatang:
:+1:

PR #581, opened 2016-03-14
check minimum dm version

### LINKED ISSUE #565: check that `docker-machine version` is a minimum
as we do with `docker version`. just upgraded from 0.5.6 to 0.6.0 & it solved a lot of problems. should also match what we're testing in circle

zramsay:
closes #565


zramsay:
this feature is skipped by appveyor because of known issue (cc @pietv)  with installing docker-machine > 0.4.1

PR #583, opened 2016-03-14
mkdir on data import if not existing

### LINKED ISSUE #501: data import should mkdir in the container if it doesn't exit
I'm on the fence. Would be ++ convenient but seems very non-unixy

zramsay:
closes #501 
requires https://github.com/eris-ltd/eris-by-example/pull/3 (a test) to pass

### DIFF
--- a/data/operate.go
+++ b/data/operate.go
@@ -28,11 +28,14 @@ import (
 //  do.Source                     - directory which should be imported (required)
 //  do.Destination                - directory to _unload_ the payload into (required)
 //
+// If the named data container does not exist, it will be created
+// If do.Destination does not exist, it will be created
 func ImportData(do *definitions.Do) error {
 	log.WithFields(log.Fields{
 		"from": do.Source,
 		"to":   do.Destination,
 	}).Debug("Importing")
+
 	if util.IsDataContainer(do.Name) {
 
 		srv := PretendToBeAService(do.Name)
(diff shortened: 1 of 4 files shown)


data/operate.go
```diff
@@ -65,19 +86,16 @@ func ImportData(do *definitions.Do) error {
...
+		// and eris images have the `eris` user by default
+		if err := runData(containerName, []string{"chown", "--recursive", "eris", do.Destination}); err != nil {
+			return err
 		}
+
 	} else {
-		log.WithField("name", do.Name).Info("Data container does not exist.")
+		log.WithField("name", do.Name).Info("Data container does not exist, creating it.")
 		ops := loaders.LoadDataDefinition(do.Name)
+		do.MakeDestination = true
```
ebuchman: whats this for?

cmd/data.go
```diff
@@ -178,6 +178,7 @@ func ImportData(cmd *cobra.Command, args []string) {
 	do.Name = args[0]
 	do.Source = args[1]
 	do.Destination = args[2]
+	do.MakeDestination = false // check if exists
```
ebuchman: dont we also need a flag?

data/operate.go
```diff
@@ -46,7 +50,24 @@ func ImportData(do *definitions.Do) error {
...
+			}
+		} else {
+			doCheck := definitions.NowDo()
+			doCheck.Name = do.Name
+			doCheck.Operations.Args = []string{"ls", do.Destination}
+			_, err := ExecData(doCheck)
+			// if an error ls-ing, try importing again
+			// but by making the dir first
+			if err != nil {
+				do.MakeDestination = true
```
ebuchman: whats the point of MakeDestination if we're just gonna go for it anyways?

zramsay: it won't go for it here if there's no error listing...

pietv:
LGTM :+1:

PR #591, opened 2016-03-16
Added CORS handler (github.com/rs/cors)

j-h-scheufen:
Minor change in agent to allow cross origin requests


agent/operate.go
```diff
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/rs/cors"
```
zramsay: this needs to be godepped

Issue #607, opened 2016-03-22
export/import all keys


zramsay:
from #356, related-ish to #597

zramsay:
closed by #610

compleatang:
:+1:

PR #610, opened 2016-03-23
import/export all keys + can use relative paths now

### LINKED ISSUE #597: relative path for `eris keys import --src`
check export as well

zramsay:
new flag to move all keys to and from container <=> host
- `eris keys import/export --all` => provide no address; uses default paths
- [x] todo: test paths properly (when I haz moar bandwidth)
- [x] add test for this new feature

closes #597 and #607


pietv:
:+1:

PR #613, opened 2016-03-25
clean: proper & prettier path logging

zramsay:
beautification of clean prompt warning


zramsay:
protip: don't edit code on GH. the PR will go to master by default

PR #632, opened 2016-04-11
agent: mini-server for installing and deploying tar'ed contract bundles

### LINKED ISSUE #520: `eris agent`
from slack:
an `agent` with an `install` endpoint. It will need to pull a tarball from ipfs, extract it, and move it into a path. if we want to be able to do this from a frontend we need something at the cli level (host level) which can receive and install (because we use the host's path for all the toml's, etc.)

zramsay:
- This PR introduces a new command: `agent` (closes #520)
- It has a single sub-command: `eris agent start` which will launch a server on `localhost:17552`.
- The server exposes three endpoints useful for download and installing (deploying) versioned contract bundles.
- These bundles are tarballs that have been added to IPFS. Support for direct tarball installs is forthcoming (#646).

#### chains
- The `/chains` endpoint (using `GET`) will list the running chains on the host where the agent was started.
-  For example, if three chains named `alice, bob, charlie`, each started via the `eris chains new` command respectively, then (once the agent has been started):
 `curl -X GET http://localhost:17552/chains` will return 
`[{ name: 'alice' }, { name: 'bob' }, { name: 'charlie' }]`
to the caller.

#### download
- The `/download` endpoint (`POST`) requires four arguments: 
`groupId, bundleId, version, hash` 
- The first three specify the path using the [Apache Maven](https://maven.apache.org/) template, while `hash` specifies an IPFS hash of the tarball to download and unpack. **note:** see below for tarball specification.
- No chain is required to be running. 
- For example: `curl -X POST http://localhost:17552/download?groupId=com.erisindustries&bundleId=marmoty-contracts&version=2.9.2&hash=QmfDrmYXKqowPMNQs1sEJXCx5fKTM2wQATNUCWG9bFA1cW` will drop the tarball that is that ipfs hash into `~/.eris/bundles/com-erisindustries/marmoty-contracts/2.9.2/` and unpack the contents of the tarball in that directory.

#### install
- The `/install` endpoint (`POST`) requires all the arguments from `/download`, in addition to: `chainName, address` 
- Here, `chainName` must be a running chain found via the `/chains` endpoint, and `address` is the address to deploy from. 
- This address must 1) be available in the linked keys container running alongside the chain and 2) have deploy permission on the named chain. 
- For example: `curl -X POST http://localhost:17552/install?groupId=com.erisindustries&bundleId=marmoty-contracts&version=2.9.2&hash=QmfDrmYXKqowPMNQs1sEJXCx5fKTM2wQATNUCWG9bFA1cW&chainName=alice&address=C7A4F01D58FC60429A3330CED519BBE14563FFA4` will drop the contract bundle in `~/.eris/bundles/com-erisindustries/marmoty-contracts/2.9.2/` **and** deploy it to the named chain. This latter functionality is akin to running `eris pkgs do --chain=alice --address=C7A4F01D58FC60429A3330CED519BBE14563FFA4` from within the `.../2.9.2/` directory. Note that bundle directory (`.../2.9.2/`) will contain an `epm.json` file and the `abi/` directory after installation. 

### Tarball Specification
To construct a tarball that meets the specification for deployment via `/install`, you must use the tar command like so to construct your bundles:
```
tar -czvf addMeToIPFS.tar.gz -C someDirectory .
```
where `someDirectory` contains an `epm.yaml` and, for example, a `myContract.sol` file. See [this StackOverflow answer](http://stackoverflow.com/a/3035446/5439085) for more details on that tar command.
- **Note:** it is highly recommended that the Tarball Specification be followed for the `/download` endpoint, however, it is not critical to its functionality. The `/install` endpoint, however, requires adherence to the specification because it makes assumptions about the location of the `epm.yaml` file.

### Usage
- 1) [run this setup script](https://github.com/eris-ltd/coding/blob/master/scripts/run/run.sh) with `bash run.sh`
- 2) note the chain name from the script and the address that it outputs
```
mkdir someDirectory 
touch someDirectory/epm.yaml && touch someDirectory/idi.sol
```
where `idi.sol` looks like:
```
contract IdisContractsFTW {
  uint storedData;

  function set(uint x) {
    storedData = x;
  }

  function get() constant returns (uint retVal) {
    return storedData;
  }
}
```
and where `epm.yaml` looks like:
```
jobs:

- name: setStorageBase
  job:
    set:
      val: 5

- name: deployStorageK
  job:
    deploy:
      contract: idi.sol
      wait: true

- name: setStorage
  job:
    call:
      destination: $deployStorageK
      data: set $setStorageBase
      wait: true

- name: queryStorage
  job:
    query-contract:
      destination: $deployStorageK
      data: get

- name: assertStorage
  job:
    assert:
      key: $queryStorage
      relation: eq
      val: $setStorageBase
```
At this point, we run the tar command: `tar -czvf addMeToIPFS.tar.gz -C someDirectory .` (<= note the period, it's important!) which will output `addMeToIPFS.tar.gz` to `pwd`, then:
```
eris services start ipfs
eris files put addMeToIPFS.tar.gz
```
which will return a hash. Note that we know: a) the name of a running chain, b) an address, c) all the path information, and d) an ipfs hash of the tarball to deploy. These can now be passed into the URL as specified in the `/download` and `/install` endpoint description.


j-h-scheufen:
See comment on: https://github.com/eris-ltd/eris-cli/commit/d9fe708128dcba3e450711f29f12d58cac8d5ee4

One issue detected during testing: for both 'install' and 'download' commands, if the `hash` URL parameter is missing, the agent fails when looking for an empty IPFS hash and returns HTTP 500. It would be better to detect the missing parameter earlier and return 'BadRequest' (HTTP 400?) as in the other cases when required params are missing.

zramsay:
@j-h-scheufen: PR description has been edited to include comprehensive explanation of functionality. Please review.

PR #649, opened 2016-04-17
by default eris should not allocate containers a tty

compleatang:
This pull request fixes our logspout problem.

Glider's image does not logspout containers which are started with a tty. Since we expressly give a tty to interactive containers, and since they are the only ones which *should* get a tty, I have changed this default.


pietv:
Execs (`services exec`, `chains exec`, `data exec`) are not affected by this change, so good to merge.

PR #657, opened 2016-04-19
changes to cli to integrate in overwrite commands

VoR0220:
This is for the epm suite so that we can implement an overwriting feature for jobs, with a default set to false.

### DIFF
--- a/cmd/pkgs.go
+++ b/cmd/pkgs.go
@@ -81,6 +81,7 @@ func addPackagesFlags() {
 	packagesDo.Flags().StringVarP(&do.DefaultAddr, "address", "a", "", "default address to use; operates the same way as the [account] job, only before the epm file is ran")
 	packagesDo.Flags().StringVarP(&do.DefaultFee, "fee", "w", "1234", "default fee to use")
 	packagesDo.Flags().StringVarP(&do.DefaultAmount, "amount", "y", "9999", "default amount to use")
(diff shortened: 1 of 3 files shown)


cmd/pkgs.go
```diff
@@ -81,6 +81,7 @@ func addPackagesFlags() {
 	packagesDo.Flags().StringVarP(&do.DefaultAddr, "address", "a", "", "default address to use; operates the same way as the [account] job, only before the epm file is ran")
 	packagesDo.Flags().StringVarP(&do.DefaultFee, "fee", "w", "1234", "default fee to use")
 	packagesDo.Flags().StringVarP(&do.DefaultAmount, "amount", "y", "9999", "default amount to use")
+	packagesDo.Flags().BoolVarP(&do.Overwrite, "overwrite", "t", false, "overwrite jobs of the same name")
```
zramsay: is `-t` standard shortflag for `--overwrite` ?

VoR0220: no...I would have put ```-w``` but it was already taken.

compleatang: The default here should actually be true as epm isn't ran in interactive mode so won't actually be able to get the yes to the question when ran from cli.

zramsay:
these commits can be squashed with `git rebase -i HEAD~3`
the `dont merge` labels is also still applied.

zramsay:
LGTM

PR #671, opened 2016-04-24
Do not assume bash is installed.

AFDudley:
Needed to pass tests on alpine base images.


zramsay:
im :+1: on this, @pietv ?

pietv:
:+1:

PR #680, opened 2016-04-28
Have CircleCI clean up docker-machine instances in AWS after test failures

AFDudley:
I ran this in CircleCI but I don't know if that's sufficient as a test, I did not attempt to interrupt any tests.


AFDudley:
Oops, wrong branch.

PR #681, opened 2016-04-28
 Have all three CIs clean up docker-machine instances in AWS after test failures

AFDudley:
I ran this in CircleCI but I don't know if that's sufficient as a test, I did not attempt to interrupt any tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ install:
 
 script:
 - tests/test.sh
+- tests/dm_cleanup.sh
 
 notifications:
   email: false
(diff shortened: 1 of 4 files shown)


AFDudley:
@pietv yeah, that's one of the reasons I don't run docker-machine in eris-marmont. After we have run with this change for a while, we can revisit and add more checks, but after f4b7 is in all the repos, we shouldn't see anymore orphaned "machines".

pietv:
@AFDudley :+1: All the better! That's just a friendly heads-up! :smile:

PR #719, opened 2016-05-26
Native arm support

shuangjj:
- pull arm docker images from quay.io when underlying platform is arm by build constraint on version/images
- detect running platform and pull chain and services files for corresponding platform.
- `eris pkgs do` remote compiler 0.12.0 won't work, reduced to 0.11.3 for arm.

### DIFF
--- a/cmd/pkgs_arm.go
+++ b/cmd/pkgs_arm.go
@@ -1,3 +1,5 @@
+// Added for ARM since the issue: https://github.com/eris-ltd/eris-cli/issues/751
+// TODO: Remove after issue fixed.
 package commands
 
 import (
--- a/cmd/pkgs_arm.go
+++ b/cmd/pkgs_arm.go
@@ -1,5 +1,3 @@
-// +build !arm
-
 package commands
 
 import (
@@ -9,7 +7,7 @@ import (
 	"strings"
 
 	"github.com/eris-ltd/eris-cli/pkgs"
-	"github.com/eris-ltd/eris-cli/version"
+	_ "github.com/eris-ltd/eris-cli/version"
 
 	. "github.com/eris-ltd/common/go/common"
 
@@ -115,7 +113,7 @@ func PackagesDo(cmd *cobra.Command, args []string) {
 }
 
 func formCompilers() string {
-	verSplit := strings.Split(version.VERSION, ".")
+	verSplit := strings.Split("0.11.3", ".")
 	maj, _ := strconv.Atoi(verSplit[0])
 	min, _ := strconv.Atoi(verSplit[1])
 	pat, _ := strconv.Atoi(verSplit[2])
--- a/initialize/writers.go
+++ b/initialize/writers.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/eris-ltd/eris-cli/util"
@@ -17,9 +18,9 @@ import (
 
 	ver "github.com/eris-ltd/eris-cli/version"
 
-	log "github.com/eris-ltd/eris-logger"
 	"github.com/eris-ltd/common/go/common"
 	"github.com/eris-ltd/common/go/ipfs"
+	log "github.com/eris-ltd/eris-logger"
 	docker "github.com/fsouza/go-dockerclient"
 )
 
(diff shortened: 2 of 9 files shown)


initialize/writers.go
```diff
@@ -171,6 +172,13 @@ func drops(files []string, typ, dir, from string) error {
 	} else if typ == "chains" {
 		repo = "eris-chains"
 	}
+	// on different arch
+	archPrefix := ""
+	if runtime.GOARCH == "arm" {
+		if repo != "eris-actions" {
```
zramsay: actions shouldn't care about arch at all. this code snippet can go

version/files_arm.go
```diff
@@ -0,0 +1,36 @@
...
+
+var (
+	SERVICE_DEFINITIONS = []string{
+		//"arch/arm/bigchaindb.toml",
+		//"arch/arm/btcd.toml",
+		//"arch/arm/bitcoincore.toml",
+		//"arch/arm/bitcoinclassic.toml",
+		//"arch/arm/compilers.toml",
+		//"arch/arm/geth.toml",
+		"ipfs.toml",
```
zramsay: arch/arm/ipfs.toml ?

version/files_arm.go
```diff
@@ -0,0 +1,36 @@
...
+		"keys.toml",
```
zramsay: ibid

version/files_arm.go
```diff
@@ -0,0 +1,36 @@
...
+		//"arch/arm/openbazaar.toml",
+		//"arch/arm/rethinkdb.toml",
+		//"arch/arm/toadserver.toml",
+		//"arch/arm/tinydns.toml",
+		//"arch/arm/tor.toml",
+		//"arch/arm/watchtower.toml",
+		//"arch/arm/do_not_use.toml",
+	}
+
+	ACTION_DEFINITIONS = []string{
```
zramsay: not needed for arch

version/version_arm.go
```diff
@@ -0,0 +1,9 @@
+package version
+
+const VERSION = "0.12.0"
+const DOCKER_VER_MIN = "1.8.0"
+
+//-$- Hypriot doesn't support 0.6.0 version docker machine -$-
```
zramsay: we're gonna need to either do it ourselves or find a workaround

initialize/writers.go
zramsay: oh, missed the != (!). this is fine :+1:

zramsay:
let's split up the arm & alpine PR's. once alpine is merged & stable we'll move forward with arm (re-built with alpine base (w/o tags...) )

zramsay:
rebase? also need to strip alpine tags since all images are based on alpine now.

shuangjj:
Stripped all alpine tags for ARM.

cmd/pkgs_arm.go
```diff
@@ -0,0 +1,121 @@
...
+		IfExit(fmt.Errorf("please provide the name of a running chain with --chain"))
+	}
+	if do.DefaultAddr == "" {
+		IfExit(fmt.Errorf("please provide the address to deploy from with --address"))
+	}
+	IfExit(pkgs.RunPackage(do))
+}
+
+func formCompilers() string {
+	verSplit := strings.Split("0.11.3", ".")
```
shuangjj: "0.12.0" compiler is not working yet. Workaround here for testing only. This file will be removed after issue fixed.

zramsay: hm. this is unfortunate. any idea why 0.12.0 compilers don't work on ARM? @VoR0220

cmd/pkgs_arm.go
```diff
@@ -0,0 +1,121 @@
+package commands
```
zramsay: since this is only needed for down-versioning the compilers, let's make an issue & add the issue # to the top of this file as a TODO

zramsay: let's also put a warning at the top of `cmd/pkgs.go`

version/files_arm.go
```diff
@@ -0,0 +1,20 @@
+package version
+
```
zramsay: add deprecation warning found in `version/files.go`

cmd/pkgs_arm.go
shuangjj: https://github.com/eris-ltd/eris-cli/issues/751

cmd/pkgs_arm.go
VoR0220: If I'm tbh I have no idea, but I don't think it's just an ARM issue (though that would be interesting if it was). Either way...refactor should take care of it...hopefully.

zramsay:
Great job @shuangjj !

PR #740, opened 2016-06-05
Sort docker images

zramsay:
Although `quay.io/eris/data` is somewhat redundant with `quay.io/eris/base`, the recent switch to alpine uncovered some unwarranted dependencies, i.e., the nightly build script (`/common/docker/build.sh`) could break eris due to lack of versioning. For this reason, [versioning was added to keys](https://github.com/eris-ltd/eris-keys/pull/47/). 

To maintain stack stability, the `data` image is now versioned (and built by circle) and replaces `base` throughout the codebase. In this way, all images (save for IPFS) are now versioned and cannot be affected by the build script (except when re-built from `base` via circle, which will still be built nightly but uses a digest for even greater stability).

This PR is pre-requisite for https://github.com/eris-ltd/common/issues/50

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -40,12 +40,12 @@ deployment:
       - docker push quay.io/eris/eris
       - docs/build.sh master
       # build data image because it is versioned
-      - tests/build_data_image.sh
+      - tests/build_other_images.sh
       - docker push quay.io/eris/data
   develop:
     branch: develop
     commands:
       - docs/build.sh
       # see above
-      - tests/build_data_image.sh
+      - tests/build_other_images.sh
       - docker push quay.io/eris/data
--- a/circle.yml
+++ b/circle.yml
@@ -39,7 +39,13 @@ deployment:
       - docker rmi quay.io/eris/eris:master
       - docker push quay.io/eris/eris
       - docs/build.sh master
+      # build data image because it is versioned
+      - tests/build_data_image.sh
+      - docker push quay.io/eris/data
   develop:
     branch: develop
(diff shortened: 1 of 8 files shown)


tests/data/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM quay.io/eris/base
```
compleatang: `base:alpine`

tests/data/Dockerfile
```diff
@@ -0,0 +1,4 @@
...
+MAINTAINER Eris Industries <support@erisindustries.com>
+WORKDIR /home/$USER/.eris
+VOLUME /home/$USER/.eris
```
compleatang: can we just get this dockerfile from common with curl rather than keeping it here as it is a bit weird that we're keeping it here in cli's repo.

circle.yml
```diff
@@ -39,7 +39,13 @@ deployment:
       - docker rmi quay.io/eris/eris:master
       - docker push quay.io/eris/eris
       - docs/build.sh master
+      # build data image because it is versioned
+      - tests/build_data_image.sh
```
compleatang: wondering if this should be build_other_images.sh or something like that and we should get our ipfs dockerfile from common and build it. as IPFS starts to add more features which we may want to expose, there are likely advantages to building it with data as the rest of the images we use by default are in our more well trafficked repos.

zramsay: agree. changed. IPFS versioning in ticket #742

PR #793, opened 2016-07-06
expose chain and keys ports; closes #792

### LINKED ISSUE #792: Expose chain and keys ports to eris pkgs do as a flag
**Problem** -- currently we hardcode in the chain's RPC port and keys port in [this line](https://github.com/eris-ltd/eris-cli/blob/master/pkgs/operate.go#L138). When folks change their ports this is too opinionated.

**Solution** -- in the longer term this should just work^tm via reading the chain and keys definition files to find the ports. In the short term we should expose this as a flag for adjustment by users if they're using a non-standard port.

compleatang:


pietv:
LGTM :+1:

PR #840, opened 2016-08-01
add a local compilers flag for easy usage of local compilers image

### LINKED ISSUE #837: add a `--local-compiler` flag to eris pkgs do
**Problem** -- it is way too many commands to start the compilers locally, store the IP address and then to route requests to it. In the future when we change default to using local compilers we need a more elegant framework here anyway. This flag will be the first step in a coherent way to run compilers service locally.

**Solution** -- add a local compilers flag which will (a) start the compilers locally, (b) load the IP:port for the compilers, (c) change epm command to utilize ^.

compleatang:
Changes:

* CompilersPort added to global config (note, both it an CompilersHost currently unused but good for future work)
* Placeholder for compilers image added to globalConfig
* When `-z` or `--local-compilers` given to `eris pkgs do` the compilers service will be started (currently using the compilers service definition file, which utilizes the `:latest` image), linked to, and removed during cleanup
* added appropriate tests for this functionality

closes #837

### DIFF
--- a/cmd/pkgs.go
+++ b/cmd/pkgs.go
@@ -59,6 +59,7 @@ func addPackagesFlags() {
 	packagesDo.Flags().StringVarP(&do.ChainPort, "chain-port", "", "46657", "chain rpc port")
 	packagesDo.Flags().StringVarP(&do.KeysPort, "keys-port", "", "4767", "port for keys server")
 	packagesDo.Flags().BoolVarP(&do.Overwrite, "overwrite", "t", true, "overwrite jobs of the same name")
+	packagesDo.Flags().BoolVarP(&do.LocalCompiler, "local-compiler", "z", false, "use a local compiler service; overwrites anything added to compilers flag")
 }
 
 func PackagesDo(cmd *cobra.Command, args []string) {
@@ -71,7 +72,7 @@ func PackagesDo(cmd *cobra.Command, args []string) {
 	if do.ChainName == "" {
 		IfExit(fmt.Errorf("please provide the name of a running chain with --chain"))
 	}
-	if do.DefaultAddr == "" {
+	if do.DefaultAddr == "" { // note that this is not strictly necessary since the addr can be set in the epm.yaml.
 		IfExit(fmt.Errorf("please provide the address to deploy from with --address"))
 	}
 	IfExit(pkgs.RunPackage(do))
--- a/config/config.go
+++ b/config/config.go
@@ -6,14 +6,16 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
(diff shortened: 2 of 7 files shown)


cmd/pkgs.go
```diff
@@ -71,7 +72,7 @@ func PackagesDo(cmd *cobra.Command, args []string) {
 	if do.ChainName == "" {
 		IfExit(fmt.Errorf("please provide the name of a running chain with --chain"))
 	}
-	if do.DefaultAddr == "" {
+	if do.DefaultAddr == "" { // note that this is not strictly necessary since the addr can be set in the epm.yaml.
```
zramsay: so should epm expose the address check then (b/c it needs to be checked somewhere)?

config/config.go
```diff
@@ -111,18 +114,31 @@ func LoadGlobalConfig() (*viper.Viper, error) {
...
-	// image defaults...
+
+	// compilers defaults
+	globalConfig.SetDefault("CompilersHost", "https://compilers.eris.industries")
+	verSplit := strings.Split(ver.VERSION, "-")
+	verSplit = strings.Split(verSplit[0], ".")
+	maj, _ := strconv.Atoi(verSplit[0])
+	min, _ := strconv.Atoi(verSplit[1])
+	pat, _ := strconv.Atoi(verSplit[2])
+	globalConfig.SetDefault("CompilersPort", fmt.Sprintf("1%01d%02d%01d", maj, min, pat))
```
zramsay: :) gonna make a lot of lives much easier

pkgs/operate.go
```diff
@@ -237,12 +246,23 @@ func CleanUp(do *definitions.Do, pkg *definitions.Package) error {
 		log.Debug("No throwaway chain to destroy")
 	}
 
+	// removal of local compiler; [csk] note we may not want to remove the container for performance reasons
+	if do.LocalCompiler {
+		log.Debug("Turning off and removing local compiler container")
+		doStop := definitions.NowDo()
+		doStop.Operations.Args = []string{"compilers"}
+		doStop.Rm, doStop.Force, doStop.RmD, doStop.Volumes = true, true, true, true
```
zramsay: we need to do this more often, rather than 4 lines of code...

zramsay:
LGTM. works well. tested a few times (and with WiFi turned off). only thing i noticed is that compiling takes quite a bit longer....to the point where i thought there was a problem & killed the process

cmd/pkgs.go
compleatang: epm won't know about that and probably shouldn't. better way here is to ship the checkedout key feature and then pass that in, imv.

compleatang:
Rebased.

PR #894, opened 2016-08-23
deprecate `eris chains new` and merge with `eris chains start`

### LINKED ISSUE #651: [Chains] eris chains new => eris chains start && some other cleanup of chains
**Problem** -- there is some cognitive overhead around "how we make the files I need to turn the thing on". For the services stream this is usually `eris services new`. For the chains stream this is usually `eris chains make`. For the upcoming remotes command this should be `eris remotes make`. 

**Solution** -- Two steps we could do to simplify this. One of which I'm keen to do, the other I'd like feedback. 

First, we should move `eris services new` -> `eris services make`. Since new on chains does something that goes vroom, new on services just drops some files, and eris chains make just drops some files, the first way we should harmonize is just to change services new to services make. 

Second, and the one I'm not so sure about, is to then collapse the chains new|start into one. Right now there is significant overlap between these two which could potentially be reduced. That said, there is the importing data part of chains new which is different and unique from how we do services (where we don't assume there will be a data import). All in all, I'm modestly against collapsing new and start for chains, because I do think these are distinct enough to handle the cognitive overhead of two commands vis a vis one.

zramsay:
Significant refactor of the `eris chains new/start` sequence (closes #651).

- significant removal of unused chains code
- `eris chains new --dir` is replaced by `eris chains start --init-dir` and `eris chains new` is completely deprecated
- `eris chains make newChain` now assumes `--chain-type=simplechain` rather than dropping user into the wizard. The wizard can be invoked with `eris chains make someChain --wizard`
- currently blocking on 1) some final cleanup of code, 2) squash commits, 3) reset feature branches pointers in circle and 4) an awkward failing tests in each epm and ecm
- merge to happen alongside identically named branches in each epm and ecm because of porcelain `eris` tests

- failing tests: 
  - https://circleci.com/gh/eris-ltd/eris-cm/218
  - https://circleci.com/gh/eris-ltd/eris-pm/561 (this one has a "try" commit on that branch)


zramsay:
closing. to be replaced by `feature-sort-configs-and-loaders` or some better-named branch that is currently branched from this branch and has the important fixes vis-a-vis the eris-cm config file situation. Note to [zr]: re-use PR description in follow-up PR

PR #898, opened 2016-08-25
Cm image respect

compleatang:
changes the usage of images so that what is populated in eris.toml is respected for pm and cm (also defaults on chains, which are currently overwritten by default.toml in ~/.eris/chains, but @zramsay is fixing that)


zramsay:
LGTM. tests failing for other reasons soon to be dealt with.

PR #902, opened 2016-08-26
DO NOT MERGE; contemplated changes necessary to use cm as a lib in cli

compleatang:


zramsay:
replaced by #1072

zramsay:
BTW, this was super helpful getting the initial integration working :)

Issue #577, opened 2016-03-13
Bug in services rm --file


compleatang:
```
$ eris services rm --file idi                                                                                                                                                           Removing file                            file=/home/coda/.eris/services/home/coda/.eris/services/idi.toml.toml
remove /home/coda/.eris/services/home/coda/.eris/services/idi.toml.toml: no such file or directory
```

Somehow we have double tomls and double services dirs.

zramsay:
closed by #580

PR #911, opened 2016-09-03
Implements one click eris startup and shutdown of services (Issue #766)

### LINKED ISSUE #766: [Core] One click shutdown/startup for dev machines
Can we have a one click roll up/down? Like `eris startup` `eris shutdown`? It would be handy for dev machines where you want to end a marmoty day of dev and gracefully shutdown everything and return the next day in the same state. I think that it would just wrap `eris services stop xxxxx` and `eris chains stop xxxx` and vice-versa against all running containers.

slowtokyo:

- [ check ] `gofmt -s` your code
- [ check ] `git rebase develop` while on your feature/fix branch
- [ check ] use a descriptive title
- [ closes #766 ] in the comments section, include the issue # (i.e., closes #577)
- [ Adds the ability to `eris startup` which starts up all existing services that you have. `startup` also takes a `--chain` flag to simultaneously startup a chain if you like. Adds ability to `eris shutdown` which gracefully shuts down all chains and services currently running. Adds function to `KillChains` which is the same functionality as the `KillChain` command except it takes an array of chain container names in case multiple chains are running. This is similar to the behaviour of `KillServices`. ] provide a significant summary of the important changes the PR yields. Describe under-the-hood behaviour of code changes and/or modifications to the user experience

### DIFF
--- a/chains/operate.go
+++ b/chains/operate.go
@@ -43,6 +43,46 @@ func InstallChain(do *definitions.Do) error {
 	return setupChain(do, loaders.ErisChainInstall)
 }
 
+//function for killing multiple chains (takes array)
+func KillChains(do *definitions.Do) error {
+	var chains []*definitions.Chain
+
+	log.WithField("args", do.Operations.Args).Info("Building chains group")
+	for _, chainName := range do.Operations.Args {
+		c, e := BuildChainsGroup(chainName)
+		if e != nil {
+			return e
+		}
+		chains = append(chains, c...)
+	}
+	
+	// if force flag given, this will override any timeout flag
+	if do.Force {
+		do.Timeout = 0
+	}
+
+	for _, chain := range chains {
+		if util.IsChain(chain.Name, true) {
+			log.WithField("=>", chain.Name).Debug("Stopping chain")
+			if err := perform.DockerStop(chain.Service, chain.Operations, do.Timeout); err != nil {
+				return err
+			}
+
+		} else {
+			log.WithField("=>", chain.Name).Info("Chain not currently running. Skipping")
+		}
+
+		if do.Rm {
+			if err := perform.DockerRemove(chain.Service, chain.Operations, do.RmD, do.Volumes, do.Force); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+//function for killing a single chain
 func KillChain(do *definitions.Do) error {
 	chain, err := loaders.LoadChainDefinition(do.Name)
 	if err != nil {
@@ -510,4 +550,27 @@ func CleanUp(do *definitions.Do) error {
 	}
 
 	return nil
+}
+
+func BuildChainsGroup(chainName string, chains ...*definitions.Chain) ([]*definitions.Chain, error) {
+	log.WithFields(log.Fields{
+		"=>":        chainName,
+		"chains#": len(chains),
+	}).Debug("Building chains group for")
+	chain, err := loaders.LoadChainDefinition(chainName)
+	if err != nil {
+		return nil, err
+	}
+	if chain.Dependencies != nil {
+		for _, cName := range chain.Dependencies.Chains {
+			log.WithField("=>", cName).Debug("Found chain dependency")
+			c, e := BuildChainsGroup(cName)
+			if e != nil {
+				return nil, e
+			}
+			chains = append(chains, c...)
+		}
+	}
+	chains = append(chains, chain)
+	return chains, nil
 }
--- a/cmd/start_stop.go
+++ b/cmd/start_stop.go
@@ -1,0 +1,36 @@
+package commands
+
+import (
+	//"github.com/eris-ltd/eris-cli/definitions"
+	"github.com/eris-ltd/eris-cli/kaihei"
+
+	"github.com/spf13/cobra"
+
+	. "github.com/eris-ltd/common/go/common"
+)
+
+var Start = &cobra.Command{}
+
+var Stop = &cobra.Command{}
+
+func buildStartCommand() {
+	addStartFlags()
+}
+
+func addStartFlags() {
+}
+
+func buildStopCommand() {
+	addStopFlags()
+}
+
+func addStopFlags() {
+}
+
+func StartEris(cmd *cobra.Command, args []string) {
+	IfExit(kaihei.StartUpEris(do))
+}
+
+func StopEris(cmd *cobra.Command, args []string) {
+	IfExit(kaihei.ShutUpEris(do))
+}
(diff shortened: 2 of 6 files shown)


zramsay:
Thanks for the PR @slowtokyo :)
Unfortunately, this will have to wait until after the current release cycle - the codebase around chains (and everything else) is too much in flux at the moment. 
- [ ] Note also that we'll want to add in the tests I wrote back when we started. (https://github.com/eris-ltd/eris-cli/commit/1a6ddf4181999edd374b9dcc176a164d06dc826e)

chains/operate.go
```diff
@@ -43,6 +43,46 @@ func InstallChain(do *definitions.Do) error {
 	return setupChain(do, loaders.ErisChainInstall)
 }
 
+//function for killing multiple chains (takes array)
+func KillChains(do *definitions.Do) error {
```
zramsay: I'm not sure we want an additional function here.

cmd/start_stop.go
```diff
@@ -0,0 +1,58 @@
+package commands
+
+import (
+	//"github.com/eris-ltd/eris-cli/definitions"
```
zramsay: this line should be completely removed

cmd/start_stop.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/eris-ltd/eris-cli/kaihei"
```
zramsay: was a good name as a placeholder but we'll want something more relevant.

definitions/chains.go
```diff
@@ -7,6 +7,7 @@ type Chain struct {
 	ChainID string `mapstructure:"chain_id" json:"chain_id" yaml:"chain_id" toml:"chain_id"`
 	// type of the chain
 	ChainType string `mapstructure:"chain_type" json:"chain_type" yaml:"chain_type" toml:"chain_type"`
+	Chain string `json:"chain,omitempty" yaml:"chain,omitempty" toml:"chain,omitempty"`
```
zramsay: can we get by without an additional variable?

kaihei/start.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/eris-ltd/eris-cli/util"
+	srv "github.com/eris-ltd/eris-cli/services"
+	"github.com/eris-ltd/eris-cli/definitions"
+)
+
+func StartUpEris(do *definitions.Do) error {
+
+	// start chain
+	// doChain.Name - name of the chain (optional)
+	if do.ChainName != ""{
```
zramsay: gofmt

kaihei/start.go
```diff
@@ -0,0 +1,98 @@
...
+		doChain := definitions.NowDo()
+		doChain.Name = do.ChainName
+
+		fmt.Println("starting up your chain...")
+		if err := chains.StartChain(doChain); err != nil {
+			return err
+		}
+	}
+	
+	fmt.Println("starting up your services...")
```
zramsay: `log.Warn()`

kaihei/start.go
```diff
@@ -0,0 +1,98 @@
...
+	if len(listOfServices) == 0 {
+		return fmt.Errorf("no existing services to start")
+	}
+
+	names := make([]string, len(listOfServices))
+	for i, serviceName := range listOfServices {
+		names[i] = serviceName.ShortName
+	}
+
+	fmt.Println(names)
```
zramsay: remove?

kaihei/start.go
```diff
@@ -0,0 +1,98 @@
...
+	doStart := definitions.NowDo()
+	doStart.ServicesSlice = names
+	if err := srv.StartService(doStart); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func ShutUpEris(do *definitions.Do) error {
```
zramsay: name choice?

kaihei/start.go
```diff
@@ -0,0 +1,98 @@
...
+
+	fmt.Println("shutting down your services...")
```
zramsay: log.Warn()

kaihei/start.go
```diff
@@ -0,0 +1,98 @@
...
+	if len(listOfServices) == 0 {
+		return fmt.Errorf("no existing services to stop")
+	}
+
+	names := make([]string, len(listOfServices))
+	for i, serviceName := range listOfServices {
+		names[i] = serviceName.ShortName
+	}
+
+	fmt.Println(names)
```
zramsay: remove?

services/operate.go
```diff
@@ -79,7 +79,7 @@ func KillService(do *definitions.Do) (err error) {
 		}
 		services = append(services, s...)
 	}
-
+
```
zramsay: gofmt

chains/operate.go
slowtokyo: Yah, I wasn't sure whether or not we wanted that. I wasn't sure whether that was a dependency for other functions so created a new one instead. The question that needs answering is: _Are there cases where you would want to only stop one chain and leave others running or do you want `eris chains stop` to stop all chains all the time?_

definitions/chains.go
slowtokyo: Yes, I think so. Meant to delete that.

kaihei/start.go
slowtokyo: Sorry, remove the whole 15 lines or so?

kaihei/start.go
slowtokyo: haha, that was your bag man. Can think of something else. I like `ShutUpEris` though

kaihei/start.go
slowtokyo: remove what?

kaihei/start.go
zramsay: line 40. if we want logging, then see above note about the `eris-logger` package

zramsay: (`fmt.Println()` or `Printf()` is great for debugging, however....i sometimes leave these lying around as well.)

kaihei/start.go
zramsay: `ShutDownEris` maybe. it's not like eris is being noisy just by having a chain and some services running ;)

kaihei/start.go
zramsay: fmt.Println. Or use `eris-logger` alongside the `WithFields` parameter, if appropriate

zramsay:
@slowtokyo thanks again for the PR. Unfortunately, we'll be closing for now as we are moving in the direction of "trimming the fat" and reducing the number of commands for eris-cli. I would recommend sticking with the [two commands here](https://github.com/eris-ltd/eris-cli/issues/766#issuecomment-230478836) in a custom startup/shutdown script.

PR #913, opened 2016-09-04
remove actions & clean up init (remove latent toadserver functionality)

### LINKED ISSUE #909: [Actions] rm
**Problem** Actions are not used. They are not at all documented. And therefore they are likely unnecessary. 

**Solution** -- remove them out of the codebase. if we have a compelling reason to put them back in later we can do so at that time.

zramsay:
- closes #909
- related to https://github.com/eris-ltd/common/pull/75 and https://github.com/eris-ltd/docs.erisindustries.com/pull/187


initialize/initialize_test.go
```diff
@@ -76,84 +67,34 @@ func TestDropChainDefaults(t *testing.T) {
...
-		ifExit(err)
-	}
-
 	if err := os.MkdirAll(dirGit, 0777); err != nil {
 		ifExit(err)
 	}
 
 	switch kind {
 	case "services":
-		//pull from toadserver
```
pietv: Looks much better without having to switch to and from "rawgit".

initialize/init.go
```diff
@@ -41,8 +41,8 @@ func Initialize(do *definitions.Do) error {
 		}
 	}
 
-	//drops: services, & chain defaults from toadserver
-	log.Warn("Initializing default service, action, and chain files")
+	//drops: service definition defaults
+	log.Warn("Initializing default service defintion files")
```
pietv: defintion

pietv:
i guess `initialize/default_actions.go` didn't make into a commit.

pietv:
Let's yank `definitions/action.go` too? Or is this intentional to leave them?

pietv:
:+1: LGTM

initialize/initialize_test.go
zramsay: :)

zramsay:
good catches on the extra two files. thanks

PR #914, opened 2016-09-05
Pull selected services definitions for testing

### LINKED ISSUE #867: [Tests] remove `eris init`'s and `eris clean`'s from tests_tooltests_jenkins
**Problem** -- our unit tests can be more cleanly isolated. The test runner for those jobs should be as minimalistic as possible. Indeed, it should mostly be controlling `go tests` and that's it. 

**Solutions** -- a good start will be removing the eris inits and eris cleans from the file noted above. We'll need some deeper refactors as well.

pietv:
* `tests.TestsInit()` takes additional arguments to pull selected services:
  
  ```
  ...
  tests.TestsInit(ConnectAndPull, "keys", "ipfs", "do_not_use")
  ...
  ``` 
* Renames `files.go` and `files_arm.go` definitions not to include the `.toml` suffix (looks better/dries out both `tests.TestsInit()` function and the definitions list).
* Includes "ipfs" and "keys" back into `SERVICE_DEFINITION` to optimize pulling of the custom list.
* Partially resolves #867.

### DIFF
--- a/initialize/initialize_test.go
+++ b/initialize/initialize_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/eris-ltd/eris-cli/config"
 	"github.com/eris-ltd/eris-cli/util"
+	ver "github.com/eris-ltd/eris-cli/version"
 
 	"github.com/eris-ltd/common/go/common"
 	"github.com/eris-ltd/common/go/ipfs"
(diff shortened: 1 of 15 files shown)


initialize/initialize_test.go
```diff
@@ -97,12 +98,12 @@ func testDrops(dir, kind string) error {
 	case "services":
 		//pull from toadserver
 		if toadUp {
-			if err := dropServiceDefaults(dirToad, "toadserver"); err != nil {
+			if err := dropServiceDefaults(dirToad, "toadserver", ver.SERVICE_DEFINITIONS); err != nil {
```
zramsay: all references to toadserver can be removed. that's (mostly) handled in #913 I'll rebase that PR when this oneis merged.

services/services_test.go
```diff
@@ -24,7 +24,7 @@ func TestMain(m *testing.M) {
 	// log.SetLevel(log.InfoLevel)
 	// log.SetLevel(log.DebugLevel)
 
-	tests.IfExit(tests.TestsInit(tests.ConnectAndPull))
+	tests.IfExit(tests.TestsInit(tests.ConnectAndPull, "do_not_use", "keys", "ipfs"))
```
zramsay: should we eliminate `do_not_use` ?

zramsay:
LGTM. comments do not affect PR in it's current form

services/services_test.go
pietv: didn't know what it's used for in the first place. yeah, let's drop it!

zramsay: used for testing....added to #917

Issue #910, opened 2016-09-03
[Services] requiring a network connection shutdown without throwing an error


slowtokyo:
Please include in your bug report:
- [1.11.1 ] `docker version`
- [  0.12.0-rc3] `eris version`
- [ go1.7 ] `go version` (if applicable)
- [ 0.7.0] `docker-machine version` (if applicable / !linux)
- [ osx] operating system details (osx/windows/linux)
- [ when starting up a service like logspout that requires an internet connection the service starts and then immediately shuts down when it can't find a network connection. In this case an error explaining the situation should be thrown. ] what you expected to happen
- [ It looks like the service started without a problem but then when checking with `eris ls` the service is not actually running. ] what actually happened
- [ try running `eris services start logspout` without an internet connection ] steps to reproduction

zramsay:
thanks for the bug report @slowtokyo
This is definitely a bug which we'll address in the near-term.

slowtokyo:
Have a fix for this. Basically wait a second to see if the container times out and then throw an error informing the user that it shut down if it did time out. Will submit a PR.

zramsay:
@slowtokyo did you have a chance to finish the PR you mentioned?

slowtokyo:
No. I don't think I'll have much time for this until maybe mid-December but may be able to pick it up again after that.

zramsay:
no longer part of the direction of the platform

PR #923, opened 2016-09-08
Check network connection when starting up services and throw error if shut down

slowtokyo:

- [x] `gofmt -s` your code
- [x] `git rebase develop` while on your feature/fix branch
- [x] use a descriptive title
- [ fixes issue #910 ] in the comments section, include the issue # (i.e., closes #577)
- [ Added a check for network timeout when starting a service to make sure that the service is still running after 1 second. ] provide a significant summary of the important changes the PR yields. Describe under-the-hood behaviour of code changes and/or modifications to the user experience

### DIFF
--- a/perform/perform.go
+++ b/perform/perform.go
@@ -13,8 +13,8 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"unicode"
 	"time"
+	"unicode"
 
 	"github.com/eris-ltd/eris-cli/config"
 	def "github.com/eris-ltd/eris-cli/definitions"
@@ -252,7 +252,7 @@ func DockerRunService(srv *def.Service, ops *def.Operation) error {
 
 	log.WithField("=>", optsServ.Name).Info("Container started")
 
-	duration := time.Duration(1)*time.Second
+	duration := time.Duration(1) * time.Second
 
 	time.Sleep(duration)
 
--- a/perform/perform.go
+++ b/perform/perform.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"unicode"
+	"time"
 
 	"github.com/eris-ltd/eris-cli/config"
 	def "github.com/eris-ltd/eris-cli/definitions"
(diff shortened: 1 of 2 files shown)


pietv:
@slowtokyo thank you very much for your PR! Very much appreciate it.

The 2 problems with the solution is that it's indefinite whether the service will be running after the arbitrary one second duration and that it delays container creation for 1 second for every single service start.

It might be useful as an optional check for services that require this. Would you be able to refactor your solution to make the post-start check and the timeout optional via the `do.Operation` struct fields — that will make them loadable from the service configuration files. Please make them off my default.

Thanks very much!

perform/perform.go
```diff
@@ -251,7 +251,15 @@ func DockerRunService(srv *def.Service, ops *def.Operation) error {
...
 	log.WithField("=>", optsServ.Name).Info("Container started")
 
-	return nil
+	runSuccess := ContainerRunning(ops.SrvContainerName)
+
+	if runSuccess {
+		return nil
+	} else {
+		log.WithField("=>", ops.SrvContainerName)
+		fmt.Println("Oops! Container shut down unexpectedly. This may be a problem with your environment. Please check network connection.")
```
pietv:
```
log.WithField("=>", ops.SrvContainerName).Warn("Oops! Container shut down unexpectedly. Please check container logs")
```
`log.Warn()` and you have a period at the end of the sentence (see [CONTRIBUTING](https://github.com/eris-ltd/eris-cli/blob/master/.github/CONTRIBUTING.md#log-messages)).

The problem with the failed container could be something else other than network problems.

pietv:
Also if you don't intend to override the `err` value returned by the `DockerRunService` function and want just the heads-up to the user, you may want to wrap the `Sleep` inside a goroutine so that it doesn't delay the service creation. Thanks.

pietv:
As an alternative solution to #910 I might suggest a precondition for running specific services like if the service expects certain network ports to be opened those checks might be done before the container start. The precondition implemented as a field, for example, in the service definition file.

PR #949, opened 2016-09-25
Refactor not to use `do.Result` to return data or errors

### LINKED ISSUE #912: [Core] Refactor error handling with do.Result
* Get rid of setting and passing `do.Result` when returning errors.

pietv:
* Refactor `do.Result` to use `fmt.Errorf` to return errors or proper types to return data results.
* Clean up `eris --help` and `eris man` screens.
* Change hardcoded `~/.eris` and `$HOME/.eris` to uses `common.ErisRoot` related paths.
* Closes #912.


keys/readers.go
```diff
@@ -7,32 +7,33 @@ import (
...
-				hostAddrs = append(hostAddrs[:0], hostAddrs[1:]...)
-				for _, addr := range hostAddrs {
-					log.WithField("=>", addr).Warn()
+				// First key.
+				log.WithField("=>", result[0]).Warn("The keys on your host kind marmot")
+				// Subsequent keys.
+				if len(result) > 1 {
+					for _, addr := range result[1:] {
+						log.WithField("=>", addr).Warn()
+					}
```
zramsay: ++ much cleaner

util/clean.go
```diff
@@ -214,7 +214,3 @@ func canWeRemove(toClean map[string]bool) bool {
 	log.Warn("Authorization not given, exiting")
 	return false
 }
-
-func TrimString(strang string) string {
-	return strings.TrimSpace(strings.Trim(strang, "\n"))
-}
```
zramsay: ++ that was a naive way to do things...

zramsay:
LGTM, much cleaner :)

PR #1042, opened 2016-10-15
Check imported directory exists

pietv:
* Additional check for `data.ImportData()`.
* `common`: remove `MigrateDirs` and similar directories because they are not updated on `ChangeErisDir()`.
* `common`: update downstream (was out of date).
* Move `~/.eris/chains/HEAD` file creation into the `eris chains checkout` command.
* Fix `eris chains checkout` command to use `util.IsChain()` to determine whether a chain exists.


zramsay:
LGTM

Issue #946, opened 2016-09-22
[Core] Refactor file naming within packages


pietv:
Refactor file names within packages to more clearly suggest what particular functionality they contain (for example: name them by function names or functionality names, or merge all of them in one file or reduce a number of functions in them to a minimum); it is sometimes takes an effort (without `grep`) to remember which one of `manage`, `operate`, `readers`, or `writers` contains a function one is looking for.

PR #1044, opened 2016-10-18
Incorporate logger and IPFS client

### LINKED ISSUE #930: [Core] harmonize naming of package imports
**Problem:** There's a lack of consistency in how imported packages are (short)named: e.g. `chns "github.com/eris-ltd/eris-cli/chains"`, among others. The short-hand for `. "github.com/eris-ltd/common"` is fine _if you already know the code_ but doesn't make it easy for new contributors to navigate.

**Solution:** harmonize

### LINKED ISSUE #1006: [Logging] Revert eris-logger to utilize a common log package hosted in cli repository
All the tools should run from here.

pietv:
* Move `eris-logger` code to `log`. Closes #1006
* Move IPFS client code to `util/ipfs.go` (Other `common` code is pending resolution of import cycles).
* Harmonization of import paths and file names. Closes #930, #946.


chains/chains.go
```diff
@@ -411,3 +578,25 @@ func whatChainStuffExists(chainName string) (bool, bool, bool, bool) {
 
 	return chainDirExists, chainConfigExists, chainDataExists, chainContainerExists
 }
+
+func exportFile(chainName string) (string, error) {
+	fileName := util.GetFileByNameAndType("chains", chainName)
+
+	return ipfs.SendToIPFS(fileName, "", "")
+}
+
```
zramsay: this function is latent code and can be removed

chains/chains.go
```diff
@@ -411,3 +578,25 @@ func whatChainStuffExists(chainName string) (bool, bool, bool, bool) {
...
+func checkKeysRunningOrStart() error {
```
zramsay: there's a generalized function `services.EnsureRunning()`  that we could use instead

chains/chains.go
pietv: Yep. Good catch. :+1: I have a list of changes I'd like to make to an IPFS client, so I'd prefer to have it as a separate ticket. The idea for this one is just to migrate _as is_.

zramsay:
LGTM :) so much cleaner. The code is much more approachable too.

PR #1053, opened 2016-10-31
properly change permissions on the imported keys

### LINKED ISSUE #1052: [Chains] On Chains Start Imported Keys are Not Set to Proper User
**Problem** -- the keys that are imported from the new start `--init-dir` are not properly changing their owner to the eris user. for rc3

compleatang:
close #1052


pietv:
I'll replace this call with assigning `doKeys.Service.User = "eris"` in the previous command — it's faster this way; also found a glitch with loading chain dependencies. Will close this PR and push a fix.

PR #1095, opened 2016-11-26
chains make: warn users if no validators created

### LINKED ISSUE #1012: [Chains] Warn users when making chains that have no validators
@j-h-scheufen commented on [Sat Mar 12 2016](https://github.com/eris-ltd/eris-cm/issues/9)

`eris chains make -h` lists same examples for account-type combinations that are not usable without further modifications, because they do not result in a usable chain configuration due to 0 validators.

```
$ eris chains make myChain --account-types=Root:1,Developer:0,Validator:0,Participant:1 -- will use the flag to make your chain named myChain (non-interactive)
$ eris chains make myChain --account-types=Root:1,Developer:0,Validator:0,Participant:1 --chain-type=simplechain -- account types trump chain types, this command will use the flags to make the chain (non-interactive)
```

The resulting genesis.json has an empty "validators" array and instantiating a chain via `eris chains new ...` using this file results in a chain that is not making any blocks.
A lot of users (me included) regularly fall into the trap that the chain seemingly was created successfully, but subsequent interactions fail and only the logs reveal that the chain is 'dead'.
### Solution Proposal
- [x] Change the help output and always have 1 root and 1 validator in the samples (in case users copy it from here)
- [ ] When validators are set to 0, output a warning that the resulting chain configuration will need additional manual work to be valid (and that for non-test scenarios at least 4 validators are needed).

---

@zramsay commented on [Thu Sep 29 2016](https://github.com/eris-ltd/eris-cm/issues/9#issuecomment-250492591)

this has been cleaned up on CLI's end with the recent refactor: 1) https://github.com/eris-ltd/eris-cli/pull/932/files#diff-b0a7e837e8946dd03dcd220c54cd6692L90 and 2) https://github.com/eris-ltd/eris-cli/pull/932/files#diff-b0a7e837e8946dd03dcd220c54cd6692R397

zramsay:
- closes #1012
- applies only to the `--account-types` flag. If we wanted to get fancy, we could check the `--chain-type` but I think it's safe to assume that users writing their own will know what they're doing.

### DIFF
--- a/cmd/chains.go
+++ b/cmd/chains.go
@@ -373,6 +373,7 @@ func MakeChain(cmd *cobra.Command, args []string) {
 
 	do.Name = args[0]
 
+	// TODO clean up this logic
 	if do.Known && (do.ChainMakeActs == "" || do.ChainMakeVals == "") {
 		cmd.Help()
 		util.IfExit(fmt.Errorf("If you are using the --known flag the --validators *and* the --accounts flags are both required"))
@@ -395,10 +396,7 @@ func MakeChain(cmd *cobra.Command, args []string) {
 	}
 
 	if do.Wizard {
-		// TODO fix
-		config.Global.InteractiveWriter = os.Stdout
-		config.Global.InteractiveErrorWriter = os.Stderr
-		do.Operations.Terminal = true
+		// TODO ... something ... ?
 	} else if len(do.AccountTypes) == 0 && do.ChainType == "" && do.ChainMakeActs == "" && do.ChainMakeVals == "" {
 		// no flags given assume simplechain
 		do.ChainType = "simplechain"
--- a/maker/maker.go
+++ b/maker/maker.go
@@ -268,6 +268,7 @@ func assembleTypesFlags(accountT []*definitions.ErisDBAccountType, do *definitio
 
 			// If the number of account types is missing,
 			// assuming 1.
+			// [zr] this should throw an error ... (bad user input)
 			num int = 1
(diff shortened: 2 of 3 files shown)


cmd/chains.go
```diff
@@ -395,10 +396,7 @@ func MakeChain(cmd *cobra.Command, args []string) {
 	}
 
 	if do.Wizard {
-		// TODO fix
-		config.Global.InteractiveWriter = os.Stdout
-		config.Global.InteractiveErrorWriter = os.Stderr
-		do.Operations.Terminal = true
+		// TODO ... something ... ?
```
VoR0220: What is supposed to go here?

maker/maker.go
```diff
@@ -268,6 +268,7 @@ func assembleTypesFlags(accountT []*definitions.ErisDBAccountType, do *definitio
 
 			// If the number of account types is missing,
 			// assuming 1.
+			// [zr] this should throw an error ... (bad user input)
```
VoR0220: then throw the error?

zramsay: like if `eris chains make whatever --account-types=Full:,Root:1` then it won't error for the empty value, but instead sets it to 1

VoR0220: got it. So then maybe this comment should be doctored a bit?

cmd/chains.go
zramsay: probably nothing. will deal  in another PR

PR #1096, opened 2016-11-26
chains make: harmonize paths for single/multiple accounts

### LINKED ISSUE #1086: [chains] of exactly 1 account should still use a full path rather than the chains dir as root dir
**Problem:** It's confusing that any chain with 1 account puts all the files in `~/.eris/chains/myChain` rather than `~/.eris/chains/myChain/myChain_full_000`.

**Solution:** always use a full path.

zramsay:
- closes #1086 
- updates current test script to work and pass (`./tests/test_chains_make.sh`); proper plumbing for tests still required.
- docs/tutorials/helpers will need updating.


VoR0220:
LGTM for the most part.

Issue #1003, opened 2016-10-13
[Services/maker] init should generate the necessary default definition files from go templates


compleatang:
@zramsay commented on [Tue Oct 04 2016](https://github.com/eris-ltd/eris-services/issues/55)
- [x] remove third party service definition files & Dockerfiles
- [x] keep anything ops related for cli but move to `/hello-eris/dockerfiles`
- [x] have cli generate service definition files & chain/account-type (from cm) using go templates

zramsay:
closed by #1214

PR #1097, opened 2016-11-27
chains make default: add more chain-types other than simplechain

### LINKED ISSUE #931: [Chains] eris init should drop > 1 chain-type
**Problem:** I want to easily test !simplechains but there's a small amount of cognitive overhead each time when thinking of some nice combo of `account-types` when doing so.

**Solution:** drop some additional `complexchain.toml` in `~/.eris/chains/chain-types/` with various (useful/suggested) combinations of account-types

zramsay:
- closes #931 and replaces https://github.com/eris-ltd/eris-cm/pull/66
- adds several `*.toml` files in `~/.eris/chain/chain-types` for consumption by the `eris chains make --chain-type` flag
- these will become templates in #1003

### DIFF
--- a/initialize/default_chain_types.go
+++ b/initialize/default_chain_types.go
@@ -1,0 +1,134 @@
+package initialize
+
+func defaultSimpleChainType() string {
+	return `
+# This is a TOML config file.
+# For more information, see https://github.com/toml-lang/toml
+
+name = "simplechain"
+
+definition = """
+A simple chain type will build a single node chain. This chain type is useful
+for quick and easy prototyping. It should not be used for anything more than
+the most simple prototyping as it only has one node and the keys to that node
+could get lost or compromised and the chain would thereafter become useless.
+"""
+
+[account_types]
+Full = 1
+Developer = 0
+Participant = 0
+Root = 0
+Validator = 0
+
+[messenger]
+
+[manager]
+
+[consensus]
+`
+}
+
+func defaultAdminChainType() string {
+	return `
+# This is a TOML config file.
+# For more information, see https://github.com/toml-lang/toml
+
+name = "adminchain"
+
+definition = """
+An adminchain type has settings for prototyping a larger chain from a sysadmin point of view. With four Validator and three Full account_types, at minimum of five nodes must be up for consensus to happen.
+"""
+
+[account_types]
+Full = 3
+Developer = 1
+Participant = 1
+Root = 1
+Validator = 4
+
+[messenger]
+
+[manager]
+
+[consensus]
+`
+}
+
+func defaultDemoChainType() string {
+	return `
+# This is a TOML config file.
+# For more information, see https://github.com/toml-lang/toml
+
+name = "demochain"
+
+definition = """
+A demochain type has one node for setting up a simplechain but comes with additional Developer and Participant accounts for demonstrating a typical application. Use for prototyping only.
+"""
+
+[account_types]
+Full = 1
+Developer = 5
+Participant = 5
+Root = 0
+Validator = 0
+
+[messenger]
+
+[manager]
+
+[consensus]
+`
+}
+
+func defaultGoChainType() string {
+	return `
+# This is a TOML config file.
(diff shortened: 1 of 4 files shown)


initialize/default_chain_types.go
```diff
@@ -0,0 +1,134 @@
...
+# For more information, see https://github.com/toml-lang/toml
+
+name = "sprawlchain"
+
+definition = """
+A sprawlchain type has a little bit of everything. Modify as necessary for your ecosystem application. Will tolerate three nodes down.
+"""
+
+[account_types]
+Full = 5
```
compleatang: When we have multi validator chains, `Full`s will typically get in the way because they'll bond by default. The better practice imv is to use `Full = 0` for all chains which are not single node chains and just use `Root`'s that bond if needed. 

Rather that using `Full`'s we should encourage users to have cloud based validators only having validator permissions (which are hugely important, but at least they won't have admin rights over the application's state thereby lowering the security surface area).

initialize/default_chain_types.go
```diff
@@ -0,0 +1,134 @@
...
+"""
+
+[account_types]
+Full = 1
+Developer = 0
+Participant = 0
+Root = 0
+Validator = 0
+
+[messenger]
```
compleatang: to stay in line with edb's nomenclature think these should be `servers` .

the fields added here should map to eris-db's server values (if populated...)

initialize/default_chain_types.go
```diff
@@ -0,0 +1,134 @@
...
+
+[manager]
```
compleatang: we should default value this as `eris-mint` or some versioned equivalent.

the fields added here should map to eris-mint's config values (if populated...)

initialize/default_chain_types.go
```diff
@@ -0,0 +1,134 @@
...
+
+[consensus]
```
compleatang: we should default value this as `tendermint` or some versioned equivalent.

the fields added here should map to tendermint's config values (if populated...)

initialize/default_chain_types.go
```diff
@@ -0,0 +1,134 @@
...
+# For more information, see https://github.com/toml-lang/toml
+
+name = "adminchain"
+
+definition = """
+An adminchain type has settings for prototyping a larger chain from a sysadmin point of view. With four Validator and three Full account_types, at minimum of five nodes must be up for consensus to happen.
+"""
+
+[account_types]
+Full = 3
```
compleatang: Despite comment below, `Full = 3` makes sense to me here.

Let's add something about how we use this chain type to test long running chains within our CI system (and it's the one we should use... cc @silasdavis @benjaminbollen )

initialize/default_chain_types.go
```diff
@@ -0,0 +1,134 @@
...
+
+func defaultDemoChainType() string {
+	return `
+# This is a TOML config file.
+# For more information, see https://github.com/toml-lang/toml
+
+name = "demochain"
+
+definition = """
+A demochain type has one node for setting up a simplechain but comes with additional Developer and Participant accounts for demonstrating a typical application. Use for prototyping only.
```
compleatang: let's rename this. I'd like demochain to have a different configuration (added below)

initialize/default_chain_types.go
```diff
@@ -0,0 +1,134 @@
...
+Root = 3
+Validator = 5
+
+[messenger]
+
+[manager]
+
+[consensus]
+`
+}
```
compleatang:
```go
func defaultDemoChainType() string {
  return `
...
name = "demochain"

definition = """
A demo chain is useful for setting up proof of concept demonstration chains. It is a quick and easy way to have multi-validator, multi-developer, multi-participant chains set up for your application. This chain will tolerate 2 validators becoming byzantine or going off-line while still moving forward. You should utilize 7 different cloud instances and deploy one of the validator chain directories to each.
"""

[account_types]
Full = 0
Developer = 5
Participant = 20
Root = 3
Validator = 7

...
```

VoR0220:
LGTM

Issue #1100, opened 2016-11-28
[Compilers] Should only import one compilers package


VoR0220:
Per comments on https://github.com/eris-ltd/eris-cli/pull/1083

PR #1110, opened 2016-11-29
Update compilers usage 1100

VoR0220:
* solves #1100 
* requires #1108


definitions/package.go
```diff
@@ -1,5 +1,7 @@
 package definitions
 
+//TODO: Interface all the jobs, determine if they should remain in definitions or get their own package
```
VoR0220: 👍

Issue #929, opened 2016-09-13
[Core] update vendored packages / use glide


zramsay:

pietv:
"use glide" for what? why?

zramsay:
to version the dependencies in `/vendor`

pietv:
Using an external tool to populate go dependencies will prevent CLI being installed with `go get` which I very much like to keep. In my opinion, the `glide` to tool is immature and very messy. 

Being certain of versions populated in `/vendor` can be made with pulling them from their respective repositories.

zramsay:
we're going to need some form of dependency management for the issues in this epic; pm & cm moving into cli have dependencies on keys & edb for which we need to be cleanly managed.

The ideal solution discussed when we began the sprint was to have the dependencies checked-in to the repo, so that it remains `go get`-able but use `glide` for updating. @VoR0220 can you provide an overview for implementing glide in #1083

VoR0220:
@pietv it won't prevent it from being go gettable, you can still do so just by checking in the vendor and that will preserve it's `go get`-ability. But yes, as for glide, it's really a matter of properly defining what you're using. So here's the current CLI glide yaml just based off of running a `glide init` in the repo on epm consolidate. 

```
package: github.com/eris-ltd/eris-cli
import:
- package: github.com/BurntSushi/toml
- package: github.com/bugsnag/bugsnag-go
- package: github.com/docker/docker
  subpackages:
  - pkg/jsonmessage
  - pkg/term
- package: github.com/eris-ltd/common
  subpackages:
  - go/docs
- package: github.com/eris-ltd/eris-compilers
  subpackages:
  - network
  - util
- package: github.com/eris-ltd/eris-db
  subpackages:
  - client
  - client/core
  - keys
  - txs
- package: github.com/eris-ltd/eris-keys
  subpackages:
  - crypto
  - eris-keys
- package: github.com/fsouza/go-dockerclient
  subpackages:
  - external/github.com/docker/docker/pkg/archive
- package: github.com/kr/text
  subpackages:
  - colwriter
- package: github.com/oleiade/reflections
- package: github.com/pborman/uuid
- package: github.com/rs/cors
- package: github.com/serenize/snaker
- package: github.com/spf13/cobra
- package: github.com/spf13/pflag
- package: github.com/spf13/viper
- package: github.com/tcnksm/go-gitconfig
- package: github.com/tendermint/go-wire
- package: golang.org/x/sys
  subpackages:
  - unix
[...]

Issue #900, opened 2016-08-25
[Pkgs] pm code should be merged into this repository


compleatang:
**Problem** -- pm is slow as hell. And testing it properly is a nightmare in the current paradigm is a nightmare.

**Solution** -- as with cm, pm should be:
- imported as a library
- have its do struct populated -> https://github.com/eris-ltd/eris-pm/blob/master/definitions/do.go
- call `RunPackage` with the appropriate do struct.

zramsay:
closing since it's been on `unstable-develop` for awhile now and the tests generally pass with `./tests/test_epm.sh`

PR #1111, opened 2016-11-29
WIP - Begin using Glide

VoR0220:
* needs more work to figure out how to properly declare and vendor in our dependencies, currently will not install dependencies
* solves one part of checklist in consolidated epm #900 and #929


zramsay:
the dependencies need to remain checked-in but managed by the `glide.lock`

VoR0220:
agree, but we need to define those dependencies first.

zramsay:
did the circle.yml get updated?

VoR0220:
no? Shouldn't have?

PR #1113, opened 2016-11-29
Feature 900 reorganize hex

VoR0220:
* reorganizes the `hex` package into the `interpret` package. Also takes out redundant utils from abi and puts them into applicable categories in the `interpet` package. Also changes the names of everything that imported `hex` to `interpret`.

* Part of #900


interpret/number.go
```diff
@@ -25,7 +25,7 @@ func Big(num string) *big.Int {
 // BigD
 //
 // Shortcut for new(big.Int).SetBytes(...)
-func BigD(data []byte) *big.Int {
+func BytesToBig(data []byte) *big.Int {
```
zramsay: ++

zramsay:
clean. LGTM

PR #1083, opened 2016-11-16
[WIP] move eris-pm into eris-cli

zramsay:
- closes #900 
- deprecates https://github.com/eris-ltd/eris-pm

still todo:
- [ ] vendor edb & sort glide (@VoR0220)
- [ ] sort import of edb (needs https://github.com/zramsay/eris-db/commit/b30738807ae94985b6caba1cdb75f93925026977) to build (@benjaminbollen & @VoR0220 )
- [ ] sort tests (& re-add fixtures) (@zramsay)
- [ ] docker linking w/ flexibility (@zramsay)
- [ ] include the last few commits that went into epm prior to copying the repo in here (@VoR0220)
- [ ] sort getting the chainID (@zramsay) 
- [ ] address @compleatang's remaining comments (@VoR0220)
- [ ] re-organize packages/files (eg. remove `hex / `ebi`) (@VoR0220 & @pietv)
- [ ] write a benchmarker for jobs (to compare & as  baseline to improve) (@VoR0220)

PR #1119, opened 2016-12-01
pkgs: cleanup old code & prepare for docker links

zramsay:
- adds to #1083 
- `eris pkgs do --chain $(eris chains ip myChain) --address ADDR` works but assumes port 46657


pkgs/jobs/job_manager.go
```diff
@@ -76,7 +77,9 @@ func RunJobs(do *definitions.Do) error {
 			announce(job.JobName, "Call")
 			job.JobResult, job.JobVars, err = CallJob(job.Job.Call, do)
 			if len(job.JobVars) != 0 {
-				log.WithField("=>", job.JobVars).Info("Job Vars")
+				for _, theJob := range job.JobVars {
+					log.WithField("=>", fmt.Sprintf("%s,%s", theJob.Name, theJob.Value)).Info("Job Vars")
```
VoR0220: I would probably leave these out...I am close to a final result on how to handle these interface jobs...but if we're just focusing for now, then this will be fine.

pkgs/pkgs_test.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/eris-ltd/eris-cli/version"
 )
 
+// [zr] there's a ton of refactoring to do here!
```
VoR0220: I may just rewrite these...

pkgs/run_package.go
```diff
@@ -6,24 +6,31 @@ import (
...
+	//	return err
+	//}
+	//fmt.Println(chainIP)
 
 	// Populates chainID from the chain
-	// TODO link properly
-	keepName := do.ChainName
-	do.ChainName = fmt.Sprintf("tcp://%s:%s", do.ChainName, do.ChainPort)
-	if err = util.GetChainID(do); err != nil {
+	// TODO link properly & get chainID not from chainName
```
VoR0220: This is going to be VERY important for package management. Chain ID is a huge part of the spec that we're going to be working on. In fact I think it will be our largest contribution to the spec.

pkgs/run_package.go
```diff
@@ -6,24 +6,31 @@ import (
...
+	// XXX temp hack :-1:
+	//do.ChainID = do.ChainName
+
+	// TODO implement do.ChainURL
```
VoR0220: This should probably be run in a separate package...something for facilitating communication between chain and epm? These are just my instincts talking and feel free to disagree.

pkgs/run_package.go
```diff
@@ -32,26 +39,50 @@ func RunPackage(do *definitions.Do) error {
...
-	/*switch do.ChainName { // switch on the flag
-	case "", "$chain":
-		head, _ := util.GetHead() // checks the checkedout chain
-		if head != "" {           // used checked out chain
-			log.WithField("=>", head).Info("No chain flag or in package file. Booting chain from checked out chain")
-			err = bootChain(head, do)
-		} else { // if no chain is checked out and no --chain given, do nothing
-			log.Warn("No chain was given, please start a chain")
+	if do.LocalCompiler {
+		if err := bootCompiler(); err != nil {
```
VoR0220: ++

VoR0220:
love seeing all the fat being trimmed

pkgs/jobs/job_manager.go
zramsay: this is a fix because the old code logged the address of the point rather than any values. If it's not useful for debugging or to the user, I'm happy to remove.

pkgs/pkgs_test.go
zramsay: Please do! (they'll definitely fail since most of the functions tested are gone)

pkgs/run_package.go
zramsay: :+1: it's not overly complicated & i had it working but was janky

pkgs/run_package.go
zramsay: it's mostly encapsulated in the `func getChainIP()` below & that should cut it (once the panic is solved...)

services/services.go
```diff
@@ -481,12 +481,7 @@ func CatService(do *definitions.Do) (string, error) {
 }
 
 func InspectServiceByService(srv *definitions.Service, ops *definitions.Operation, field string) error {
-	err := perform.DockerInspect(srv, ops, field)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return perform.DockerInspect(srv, ops, field)
```
zramsay: maybe. there might be a good reason for it cc @pietv. regardless, that can happen in a later PR unrelated to getting epm in

pkgs/jobs/job_manager.go
VoR0220: Oh definitely a fine fix...for the code now...but for the code I'm planning on plopping in...won't necessarily have function. For now it can work.

PR #1122, opened 2016-12-03
agent: remove all code (currently unused)

zramsay:
- removes unused agent code (command has been commented out)
- if/when we want this feature back, it would benefit from a complete re-write anyway


VoR0220:
No no no...I was going to use this.

VoR0220:
Or atleast modify it.

zramsay:
then document that somewhere.

VoR0220:
My bad. Thought you knew bout the spec. I need a way to get IPFS hashes. And potentially deploy them...though that may come later, but if we have it right now it actually could be useful. I was actually wondering if you could help me understand IPFS better in terms of how we could structure it in terms of a repo since you seem to have actually worked deeper with it?

PR #1134, opened 2016-12-12
Feature consolidate epm chain link

zramsay:
- sorts the chain linking for `eris pkgs` (todo: flexible port/keys link)
- adds fixtures => `tests/test_epm.sh` goes green
- build still requires temp branch in eris-db with cli/log


util/paths.go
```diff
@@ -254,30 +252,3 @@ func DoesFileExist(file string) bool {
 	}
 	return true
 }
-
-// from epm
-func BundleHttpPathCorrect(do *definitions.Do) {
```
VoR0220: ++ good stuff.

VoR0220:
for now this will do. A couple of these functions I'm planning on deprecating with the stuff I'm doing in my branch. But for now these can be merged.

PR #1141, opened 2016-12-15
Feature fix urls for chains mac windows

VoR0220:
There was a problem I was running into where the docker container would time out while getting the chain ID. This is due to the fact that Linux runs with the docker host known at all times, whereas on OS X and Windows, that is made available to them through the VM of which docker is operating on. This PR takes this into account when acquiring the chain URL. Further things that need this kind of attention:

- Keys
- Possibly compilers (haven't gotten that far in the tests yet)


pkgs/run_package.go
```diff
@@ -40,20 +42,35 @@ func RunPackage(do *definitions.Do) error {
...
 		return fmt.Errorf("chain (%s) is not running", do.ChainName)
 	}
 
-	containerName := util.ContainerName(definitions.TypeChain, do.ChainName)
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+		dockerHost := os.Getenv("DOCKER_HOST")
+		port := strings.LastIndex(dockerHost, ":")
+		do.ChainIP = dockerHost[:port]
+		chainURL, err := url.Parse(do.ChainIP)
+		if err != nil {
```
pietv: there's a neater `net.SplitHostPort` function for parsing ports (https://godoc.org/net#SplitHostPort)

pkgs/run_package.go
```diff
@@ -11,12 +15,10 @@ import (
 )
 
 func RunPackage(do *definitions.Do) error {
-	if err := getChainIP(do); err != nil {
+	if err := getChainIPandURL(do); err != nil {
```
pietv: if it will be needed not just for chains (figuring out VM IP), let's move it to `util` straight away?

PR #1161, opened 2017-01-11
Updated definitions for genesis

VoR0220:
While I cannot get it to work on my machine (for the reason of keys), try this on your machine. 

In this PR I have:

- Moved maker into the chains directory (they are related and therefore makes sense to stagger them)
- Moved cm_files and cm_files_test from util into the loader package along with some others where it seemed to make sense (this is nothing permanent...no need to freak...it just bothered me the way they were currently organized)
- Fleshed out the definitions for ErisDBAccount (and there may be more in there that needs to be pruned, but now should not affect the genesis file)
- added some notes into maker
- changed an import for the dockerfile...this can be changed back but note that it will leave us as soon as we port to glide and therefore we should alter the vendor to account for this.


chains/make.go
```diff
@@ -45,7 +46,7 @@ func MakeChain(do *definitions.Do) error {
 
 	// announce.
 	log.Info("Hello! I'm the marmot who makes eris chains.")
-	keys.DaemonAddr = "http://172.17.0.2:4767" // tmp
+	keys.DaemonAddr = "http://127.0.0.1:4767" // tmp
```
zramsay: linking needed but that's a seperate issue

chains/maker/erisdb_chains.go
```diff
@@ -29,17 +29,17 @@ func MakeErisDBChain(name string, seeds []string, accounts []*definitions.ErisDB
...
+		if err := loaders.WritePrivVals(genesis.ChainID, account); err != nil {
 			return err
 		}
-		if err := util.WriteGenesisFile(genesis.ChainID, genesis, account); err != nil {
+		if err := loaders.WriteGenesisFile(genesis.ChainID, genesis, account); err != nil {
 			return err
 		}
 		// TODO: [ben] we can expose seeds to be written into the configuration file
 		// here, but currently not used and we'll overwrite the configuration file
 		// with flag or environment variable in eris-db container
```
zramsay: Lines 38-40 can go, this was implemented

chains/maker/erisdb_chains.go
```diff
@@ -29,17 +29,17 @@ func MakeErisDBChain(name string, seeds []string, accounts []*definitions.ErisDB
...
 		theSeeds := strings.Join(seeds, ",") // format for config file (if len>1)
-		if err := util.WriteConfigurationFile(genesis.ChainID, account.Name, theSeeds,
+		if err := loaders.WriteConfigurationFile(genesis.ChainID, account.Name, theSeeds,
```
zramsay: i'm not sure about the move from `util` => `loaders`. What's the reasoning exactly?

chains/maker/maker.go
```diff
@@ -18,6 +19,8 @@ var (
 )
 
 func MakeChain(do *definitions.Do) error {
+	//todo: [rj] makeRaw needs to be modified so that multiple paradigms are able to be used at once
+	// function signature should look like makeRaw(whateverReplacesDo, string...)
```
zramsay: put these comments above the `func makeRaw`

chains/maker/maker.go
```diff
@@ -58,7 +61,7 @@ func makeWizard(do *definitions.Do) error {
 		}
 	}
 
-	accountTypes, err := LoadAccountTypes()
+	accountTypes, err := loaders.LoadAccountTypes()
 	if err != nil {
```
zramsay: using `loaders` here definitely makes sense :+1:

chains/make.go
VoR0220: yea sorry I forgot to change that back. That was what I was trying to do to get it working on DFM.

definitions/account.go
```diff
@@ -1,20 +1,15 @@
...
+	Address     string                    `mapstructure:"address" json:"address"`
+	Amount      int                       `mapstructure:"amount" json:"amount"`
+	Permissions *ErisDBAccountPermissions `mapstructure:"permissions" json:"permissions"`
 
 	Validator         bool
 	PermissionsMap    map[string]int
 	ErisDBPermissions *ErisDBAccountPermissions
 	MintKey           *MintPrivValidator
+	PubKey            string
+	ToBond            int
```
zramsay: lines 9-14 still end up in the genesis.json as:
```
      "Validator": false,
      "PermissionsMap": null,
      "ErisDBPermissions": null,
      "MintKey": null,
      "PubKey": "",
      "ToBond": 0
```
although the bug is fixed, these should not be in the genesis file since they aren't needed

loaders/chains.go
```diff
@@ -55,6 +55,24 @@ func LoadChainDefinition(chainName string, definition ...string) (*definitions.C
...
+	var typ = definitions.BlankChainType()
+	var chainType = viper.New()
+
+	if err := getSetup(fileName, chainType); err != nil {
+		return nil, err
+	}
+
+	// marshall file
+	if err := chainType.Unmarshal(typ); err != nil {
+		return nil, fmt.Errorf("\nSorry, the marmots could not figure that chain types file out.\nPlease check your chain type definition file is properly formatted.\nERROR =>\t\t\t%v", err)
```
zramsay: this error message needs to be cleaned up. see https://github.com/eris-ltd/eris-cli/blob/master/.github/CONTRIBUTING.md#errors-and-log-messages-style

pkgs/run_package.go
```diff
@@ -55,6 +56,14 @@ func getChainIP(do *definitions.Do) error {
...
 	do.ChainIP = cont.NetworkSettings.IPAddress
 
+	// get the port
+	chainPortRaw := util.ParsePortMappings(cont.NetworkSettings.Ports, []string{"46657"})
+	do.ChainPort = strings.Split(chainPortRaw, ":")[1] // chainPortRaw comes out as 0.0.0.0:35487
+
+	// [zr] override & hardcode => connection to edb refused on "correct" port but contract deploy
+	// works on a chain that was started with [--publish]
+	do.ChainPort = "46657"
+
```
zramsay: this has changed. rebase required.

chains/maker/erisdb_chains.go
VoR0220: It seemed to fit better...it's "loading" the configuration file in. I guess I'm trying to follow a rule that I made in my head without telling anyone, but unless it's code that is not used by multiple packages, it should be in `util` it should find a home in its own package.

definitions/account.go
VoR0220: Yea but there were problems in terms of I believe the `priv_validator.json` file and I wasn't sure whether or not we were keeping that.

loaders/chains.go
VoR0220: this was a copy/paste. I can clean it up.

definitions/account.go
zramsay: we'll keep for now. can you find a clean way to have it not included in genesis file while maintaining status quo with priv_val ?

VoR0220: Not if priv_val continues to use ErisDBAccount. The way that it's currently structured, it's not going to get marshalled into the genesis. The better way to do things (but I'm not certain how to navigate to make this happen), would be to set aside a struct purely for priv_val.

therealplato: perhaps:
```
  PubKey string `json:"PubKey,omitempty"`
```

VoR0220: nah, we want to avoid that. But thank you for the suggestion @therealplato

chains/maker/erisdb_chains.go
VoR0220: perhaps we make a `writers` package?

definitions/account.go
VoR0220: I found one better `json:"-"`

chains/maker/erisdb_chains.go
VoR0220: I just really agree with the golang people when they say the package `util` should not exist in your stack because it doesn't describe what the package is for very well...everything should have a place and a proper name...but this is a bit philosophical I suppose.

loaders/chains.go
VoR0220: yea on second look I'm not sure how to clean this up...does logrus have something that creates errors? On first look it doesn't appear so. @pietv ?

initialize/default_keys.go
```diff
@@ -29,6 +29,7 @@ status = "unfit for production"
 [service]
 image = "` + path.Join(version.DefaultRegistry, version.ImageKeys) + `"
 data_container = true
+ports = ["4767:4767"]
```
zramsay: who now. don't think this'll fly...

writers/logging.go
```diff
@@ -0,0 +1,27 @@
+package writers
```
zramsay: this whole file could go? (probably)
IIRC it is used only by `PreRun:` in `cmd/pkgs.go`
- that PreRun should actually be removed IMO, it's unique that we're using it & an artifact of old epm
- this logging can go right in the `RunPackage` or whatever the main function is called.

writers/packages.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"reflect"
+
+	// MARMOT: dependency on go-wire needs to be removed.
```
zramsay: cut it loose :doug:

initialize/default_keys.go
VoR0220: ....how the fuck did that get in there? Thats supposed to be from my keys encapsulation....we decided we are fully exposing the ports and adding an authentication layer.

chains/make.go
```diff
@@ -6,12 +6,13 @@ import (
...
 	"strings"
 
+	"github.com/eris-ltd/eris-cli/chains/maker"
 	"github.com/eris-ltd/eris-cli/config"
 	"github.com/eris-ltd/eris-cli/definitions"
 	"github.com/eris-ltd/eris-cli/log"
-	"github.com/eris-ltd/eris-cli/maker"
 	"github.com/eris-ltd/eris-cli/services"
 	"github.com/eris-ltd/eris-cli/util"
+	"github.com/eris-ltd/eris-cli/writers"
```
pietv: @VoR0220 is it possible to refactor the `writers` package into modules the functions belong to (`chains` to chains, etc.) or at the very least move them to `util`? The functions in it are too diverse to qualify for a separate module, especially because not all of them writing anything. The reason is to simplify searching for a function you don't remember where it is defined and avoid using `grep` for it all the time; also, to avoid duplicating the `util` module purpose (there are already two: `util` and `testutil` and the latter should go too).

VoR0220: I would prefer the former solution out of all of them. Each package can have a utils file. I dont think we should be relying on a `utils` package. Every function should belong to a particular module in some form.

loaders/chains.go
pietv: @VoR0220 while it is best to keep returning simple errors and formatting them for presentation, if it is necessary, in `cmd/*.go` files, you can at least use backticks to make the newlines clearer:

```
    return nil, fmt.Errorf(`Sorry, the marmots could not figure that chain types file out.
Please check your chain type definition file is properly formatted: %v", err)
```

pietv: Could you please drop this from anywhere in the code `ERROR =>\t\t\t`? Ugly.

chains/make.go
VoR0220: Just so we're clear...are you saying put the writers into the packages for which they are writing? Or are you saying to break down the writer package into separate modules correlating to the packages for which they are writing? Because not everything in the loaders package is loading something. Some of them are util functions.

pietv: Yes, imho, it is best to keep those writing functions in packages they're writing for. 

> Because not everything in the loaders package is loading something
Yes, that's true, in which case they're at least either helping to load something or related, or they should be moved elsewhere.

chains/maker/maker.go
```diff
@@ -8,8 +8,10 @@ import (
 	"strings"
 
 	"github.com/eris-ltd/eris-cli/definitions"
+	"github.com/eris-ltd/eris-cli/loaders"
 	"github.com/eris-ltd/eris-cli/log"
 	"github.com/eris-ltd/eris-cli/util"
+	"github.com/eris-ltd/eris-cli/writers"
 )
```
pietv:
```
chains/maker/maker.go:39: unreachable code
exit status 1
pkgs/abi/abi_test.go:1101: possible formatting directive in Error call
exit status 1
util/ipfs.go:373: unreachable code
exit status 1
```

VoR0220: Yea i derped there.

VoR0220: Should go through the tests and make sure they are properly editted.

writers/logging.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"os"
+
+	"github.com/eris-ltd/eris-cli/definitions"
+	"github.com/eris-ltd/eris-cli/log"
+)
+
+// ------------------------------------------------------------------------
+// Logging
```
pietv: @VoR0220 can you put package description above the `package` keyword as is a standard practice? (https://golang.org/doc/effective_go.html#commentary) and many exported functions are missing comments too
(can you describe at least the most important ones, for now)?

VoR0220: I can indeed shoot for this. But this is quickly turning into the all in one PR.

pietv: Agree, no rush, can be in another PR(s).

loaders/chains.go
VoR0220: done.

Issue #901, opened 2016-08-26
[Keys] what do we do with it?


compleatang:
**Problem** -- running keys via an exec service is slower than it needs to be. Particularly because keys should be ran as a daemon and everything should interact with it via it's API, not its command line interface (it's made to be a server). (This is related to this epic -> https://github.com/eris-ltd/eris-keys/issues/64 )

**Solution** -- we should interact with keys the way that we currently interact with IPFS; that is to say via an API client library that we import. Note that for testing purposes this will need to be mocked.

zramsay:
bumping since keys will go in edb. This should wait until then.

zramsay:
re-named because that's really the question we need to answer first

VoR0220:
I say we create a reasonable JSON_RPC (maybe has some web3 options for the sake of portability and maybe tendermint/ABCI rpc methods?), that any client can simply implement themselves on their end for their key systems, and then they just swap in their key system. Ben's idea for this is a solid one and I back that horse. This way we can keep ours around for purely development purposes and we can get other signers on board.

zramsay:
the repo _may_ move into hyperledger/burrow, however, I don't think we should do anything else other than that

PR #1162, opened 2017-01-12
Review - Encapsulate keys

benjaminbollen:
(sorry ben, I'm editting your comment)

What this does:

* Encapsulates keys into an empty struct that can be used across the stack
* Redoes the keys test to reflect this
* Gets rid of the Do struct in the keys (except inside the function which will be fixed when data is encapsulated)
* Fixes the way that keys are used in jobs
* Fixes the way that keys are used in the chain maker
* Note: This does take note of #901, however given where we are, it will suffice for now to use the exec handler and eventually move to a traditional client/server setup in the future. 

Sincerely, 

RJ


VoR0220:
@benjaminbollen fixed it up a bit. It's now an empty struct but it works well for encapsulating methods and ensuring that one has to initialize a keys struct in order to work with the keys.

@zramsay @pietv need a review here.

initialize/default_keys.go
```diff
@@ -29,6 +29,7 @@ status = "unfit for production"
 [service]
 image = "` + path.Join(version.DefaultRegistry, version.ImageKeys) + `"
 data_container = true
+ports = ["4767:4767"]
```
zramsay: hrm

keys/keys.go
```diff
@@ -42,18 +64,18 @@ func ListKeys(do *definitions.Do) ([]string, error) {
...
 	}
 
-	if do.Container {
-		do.Name = "keys"
-		if err := services.EnsureRunning(do); err != nil {
+	if container {
+		err := keys.ensureRunning()
+		if err != nil {
 			return nil, err
 		}
```
zramsay: convention should be:
`if err := keys.ensureRunning(); err != nil {
             return nil, err
}

keys/keys.go
```diff
@@ -42,18 +64,18 @@ func ListKeys(do *definitions.Do) ([]string, error) {
...
 
-		keysOut, err := services.ExecHandler(do.Name, []string{"ls", "/home/eris/.eris/keys/data"})
+		keysOut, err := services.ExecHandler("keys", []string{"ls", "/home/eris/.eris/keys/data"})
```
zramsay: ++ i'm in favour of relying less on the do Struct

initialize/default_keys.go
VoR0220: we've been discussing this in the channel. The ports need to be published in order for it to work in the new paradigm where pm is not a container. @benjaminbollen is working on a verification system for the keys to deal with the fact that the ports for keys are now exposed.

keys/keys.go
VoR0220: ++ good suggestion.

keys/keys.go
```diff
@@ -71,84 +93,223 @@ func ListKeys(do *definitions.Do) ([]string, error) {
...
+		buf, err = services.ExecHandler("keys", []string{"eris-keys", "convert", "--name", name})
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		return nil, errors.New("Must have one and only one of either name or address.")
+	}
+
+	return buf.Bytes(), nil
+}
```
zramsay: the new functions in this file should get tests:
```
Convert
SignMsg
```

VoR0220: I didn't add tests in there because they are helper functions and the functions themselves are tested through the keys repo. But I suppose it couldn't hurt to have just one of each.

PR #1177, opened 2017-01-18
Feature implement glide

VoR0220:
* Caps off a long and arduous journey through consolidation by making dependency management slightly easier by checking in a glide.yaml, a glide.lock, and the vendor that these files created
* Let it be known here and now that the CLI repo is still go gettable and will continue to be go gettable
* Whenever the golang dep tool is created and released to the general public we can then discuss switching to it. Until such time, this is the best dependency management tool available in the go community and I fully stand by this statement.
* The process for updating a dependency is as follows:

```
glide get your/package/here
glide up
glide install
// get glide vendor cleaner plugin please...makes this thing easier to install generally
// go get -u github.com/sgotti/glide-vc
glide vc
git add vendor/*
git commit -sm "my commit message here describing the bringing in of such a dependency"
```

We are fully commited to committing in the vendor so as to continue go gettability. If this previous sentence doesn't make sense to you, don't worry about it.


zramsay:
passes on the correct jenkins job (GH integration needs to be updated once jenkins is sorted)

PR #1178, opened 2017-01-18
sort chains test WRT simplechain creation

zramsay:


pkgs/pkgs_test.go
```diff
@@ -13,12 +13,12 @@ import (
...
 	"github.com/eris-ltd/eris-cli/log"
 	"github.com/eris-ltd/eris-cli/services"
 	"github.com/eris-ltd/eris-cli/testutil"
-	"github.com/eris-ltd/eris-cli/util"
-	"github.com/eris-ltd/eris-cli/version"
+	//"github.com/eris-ltd/eris-cli/util"
+	//"github.com/eris-ltd/eris-cli/version"
 )
 
 // [zr] there's a ton of refactoring to do here!
```
zramsay: this should be handled in #1121

chains/chains_test.go
```diff
@@ -641,7 +641,7 @@ func create(t *testing.T, chain string) {
 	do := definitions.NowDo()
 	do.Name = chain
 	do.Operations.PublishAllPorts = true
-	do.Path = filepath.Join(config.ChainsPath, chain) // --init-dir
+	do.Path = filepath.Join(config.ChainsPath, chain, fmt.Sprintf("%s_full_000", chain)) // --init-dir
```
zramsay: this is required because of #1096

PR #1184, opened 2017-01-26
added self destructing contract test

VoR0220:
Two problems have been identified in this test:

1. Cannot get the account balance from a contract address...this needs to be rectified.
2. There appears to be an infinite loop on a selfdestruct command, but what's interesting is that it only manifests when it is contract->contract calling the selfdestructing contract, and doesn't manifest whenever it is account->contract calling the selfdestruct. 

@benjaminbollen @silasdavis use this test when solving these problems.


zramsay:
replaced by #1344

PR #1197, opened 2017-01-31
Linking for Binaries

VoR0220:
* enables linking for .bin files to addresses https://github.com/eris-ltd/eris/issues/1193


zramsay:
the title has "default local compilers"; why? that's what #1201 is supposed to do, right?

VoR0220:
Correct. I had them all in one and then decided it was better to split them up.

PR #1200, opened 2017-02-02
Rid Logs from EPM/EDB interaction

VoR0220:
Temporary fix for the problem of logs in EPM/eris-client relations.


zramsay:
much nicer logging (i.e., not spewing) :+1:

PR #1208, opened 2017-02-08
fix e-db configuration so that we get 0.16.0

VoR0220:
Edits the config file to use e-db 0.16.0

Signed-off-by: RJ Catalano <rj@monax.io>


benjaminbollen:
FIPF, those are duplications from what should be a dependency on e-db; where we have updated the versions.  This should also go into the release branch, because this is again needed for fixing the cross-release

VoR0220:
good point. Changing.

PR #1221, opened 2017-02-12
core/docs: remove --machine flag & begin docs consolidation

### LINKED ISSUE #1150: [Core] Remove `--machine` flag
**Problem** -- Docker machines are a very expensive abstraction for nominal (if any gain). Why?

*Costs*

* this is a docker only, not openly maintained API surface that we need to maintain. 
* the various networking layers are ridiculous (see, accessing keys for a remote docker machine)
* volume feature basically becomes ridiculous (see, our excessive use of data containers where simple volumes should do)

*Gains*

* if "feels" neat....?

**Solution** - stop supporting docker on remote machines. Benefits of this approach:

* reduced surface area on a third-party API/abstraction
* increased ability to migrate to rkt or to other container management systems

**Constraints** -- currently we need to natively support windows, osx and docker's native engines for those operating systems need to come out of beta.

zramsay:
- closes #1150, which begins the path towards un-bundling from docker-machine. Next steps include cleaning up the automatic (shelled-out) machine creation in `util/docker_client.go`, which absolutely needs to go. This will also incentivize the docs to generalize the chain setup process a bit more. 
- moves of handful of old docs to `deprecated` directory as, well, deprecated. There's been sufficient divergence to warrant a re-write of these IMO.


zramsay:
note to self: will need to remove all the docs stuff from this PR in light of doing that seperately for 0.16

zramsay:
this will be re-done with a carving knife rather than hacksaw

PR #1225, opened 2017-02-14
add release checklist from /coding repo wiki, with commit history of whole wiki

zramsay:
commit history provided to have a record of previously relevant files + how/why they were moved around.


zramsay:
@eris-ltd/team-platform please comment / PR this document to add the relevant steps for a release. I'll be adding some details provided by Peter for the cli's release. This document needs an addition from @benjaminbollen about the flow of `release-*` branches as well as details of the js libraries from @NodeGuy

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
+# eris:cli
+- [ ] Agree on a version to which the #develop branch will be bumped
```
silasdavis: Can we note that this is for the _next_ release. I'm also questioning the assumption that we will know ahead of time what the next release should be. What if we decide to do a minor instead of a patch version bump?

We still ought to bump the version number as a place-holder after releasing. But perhaps we need to make 'ensure develop has an agreed release version number' as the first step. Also can we make explicit what a version bump is (changing version.go).

In any case i think this is a little confusing for a new-comer, where it appears.

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+- [ ] draft a [PR](#draft-a-release-pr) from eris-cli:develop to eris-cli:master on Github.
+- [ ] when tests are stable, merge the PR to master
+- [ ] [tag](#tag-the-release) the release with a changelog
+- [ ] execute `misc/release/release.sh` while on the #master branch. This will:
+  * publish a release to github
+  * add the changelog from the most recent tag to github
+  * cross compile eris for various platforms
+  * builds .deb and .rpm packages and creates APT and YUM repositories
+  * upload files to Amazon S3
+- [ ] move develop to next branch
```
silasdavis: What does this mean?

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
+# eris:cli
```
silasdavis: Just `eris` ?

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+- [ ] Agree on a version to which the #develop branch will be bumped
+- [ ] execute keys's release checklist
+- [ ] execute db's release checklist
+- [ ] execute compiler's release checklist
+- [ ] execute release_script checklist
```
silasdavis: Rather than making the release of dependencies checklist items for eris-cli. Perhaps we should have an overarching checklist for 'the eris plaform'. Something like:

# The Eris Platform
A release of the Eris Platform consists of a release of eris-cli and all of its dependencies at the same version number. A release is complete when all of the artefacts for the `eris` tool and all of its dependencies have been pushed to external repositories at the same version number. The process is:

- [ ] Agree the _platform release version number_ (for the present release)
- [ ] Agree a placeholder _platform release version number_ for the next release
- [ ] Release `keys` (by following its release checklist)
- [ ] Release `db` (by following its release checklist)
- [ ] Release `compilers` (by following its release checklist)
- [ ] Release `eris` (by following its release checklist)
- [ ] Finalise the platform release by following the release_script checklist

Then for each dependency make this the first point:
- [ ] Ensure the platform release version number is referenced by the ./version/version.go version metadata file

And make this the last point:
- [ ] Bump the version number on the develop branch by referencing the placeholder platform release version number (for the next release) in the ./version/version.go metadata file

(perhaps we can come up with pithier ways of making the above two points, but I'd like to be clear)

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+- [ ] update eris' `brew` formula 
+- [ ] update eris' `choco` formula
+
+
+# eris:keys
+- [ ] [draft](#draft-a-release-pr) a PR from develop to master
+- [ ] merge develop to master
+- [ ] [tag](#tag-the-release) the release with a changelog
+- [ ] once tests pass ensure images pushed
+- [ ] version bump develop
```
silasdavis: We may want to include a final subsection (with #tag-the-release) that defines 'versioning bumping a branch'

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+  * cross compile eris for various platforms
+  * builds .deb and .rpm packages and creates APT and YUM repositories
+  * upload files to Amazon S3
+- [ ] move develop to next branch
+- [ ] version bump develop
+- [ ] update eris' `brew` formula 
+- [ ] update eris' `choco` formula
+
+
+# eris:keys
```
silasdavis: `keys`?

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+- [ ] [draft](#draft-a-release-pr) a PR from develop to master
+- [ ] merge develop to master
+- [ ] [tag](#tag-the-release) the release with a changelog
+- [ ] once tests pass ensure images pushed
+- [ ] version bump develop
+
+# eris:db
```
silasdavis: `db`

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+- [ ] [draft](#draft-a-release-pr) a PR from develop to master
+- [ ] merge develop to master
+- [ ] [tag](#tag-the-release) the release with a changelog
+- [ ] once tests pass ensure images pushed
+- [ ] version bump develop
+
+# eris:compilers
```
silasdavis: `compilers`

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+- [ ] ensure circle-ci tests against cli:master
+- [ ] [draft](#draft-a-release-pr) a PR from develop to master
+- [ ] merge develop to master
+- [ ] [tag](#tag-the-release) the release with a changelog
+- [ ] once tests pass ensure images pushed
+- [ ] version bump develop
+- [ ] start new compilers container
```
silasdavis: Is this the remote compilers service? Instructions?

docs/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+
+# eris:db
+- [ ] [draft](#draft-a-release-pr) a PR from develop to master
+- [ ] merge develop to master
+- [ ] [tag](#tag-the-release) the release with a changelog
+- [ ] once tests pass ensure images pushed
+- [ ] version bump develop
+
+# eris:compilers
+- [ ] ensure circle-ci tests against cli:master
```
silasdavis: Should this read 'ensure circle-ci tests **pass** against cli:master'? eris master is still the old version at this stage, so this doesn't seem right.

Do we effectively have a mutual dependency between compilers and eris here? If so the best we can do is make sure compiler tests pass against eris (cli) develop. We might need another step to finally make compilers depend on eris master.

Note: I have no idea how compilers are set up. Looking at https://github.com/eris-ltd/eris-compilers/blob/master/circle.yml I don't see a dependency back on eris. So I'm not sure what they refers to.

docs/release-checklist.md
VoR0220: Within the CLI there are different parts that we still refer to...Might add Readmes to each part. Trying to go for a modular approach within the CLI.

docs/release-checklist.md
VoR0220: merge it into master, that's what it should say.

docs/release-checklist.md
VoR0220: I think it's fine the way it is.

docs/release-checklist.md
VoR0220: I think a subsection is unnecessary...should probably know what a version bump is considering it's just inputting a couple of numbers.

docs/release-checklist.md
VoR0220: keep.

docs/release-checklist.md
VoR0220: keep

docs/release-checklist.md
VoR0220: Don't worry about the compilers bit in here, it's being deprecated...keep it as is for now.

docs/release-checklist.md
VoR0220: As above, don't worry about this portion, compilers is being deprecated. No new person is going to need to remember this.

zramsay:
missing info from Peter:
- 1) the gpg keys need to be put in `misc/release`. these can be regenerated but would need to be amended in `release.sh`

- 2) set these values:
```
export GITHUB_TOKEN= 
export AWS_ACCESS_KEY= 
export AWS_SECRET_ACCESS_KEY=
```
ensuring that your AWS have write access to the S3 buckets
- 3) tag the release (see existing release checklist in this PR) and docker needs to be set up (`docker-machine env`)
- 4) on master branch, run `./release.sh` which will build binaries, package, and publish them
- 5) for Chocolatey, modify the version in `misc/chocolatey/eris.nuspec` and run the `choco push .` command but you'll need an API key from their website. There's a procedure for getting one.
- 6) brew happens semi-automatically when a new release is tagged
- 7) be sure not to commit the gpg keys

zramsay:
comments addressed in #1280 (replaces this PR)

PR #1229, opened 2017-02-16
ensure pkgs test doesn't fail

zramsay:


zramsay:
@VoR0220 this isn't actually a test file:
`pkgs/jobs/jobs_test_jobs.go → pkgs/jobs/jobs_test.go`
we should re-name and have an actual test file?

VoR0220:
I have renamed it in my WIP branch for interfacing jobs. Can rename it for now, but I already have a name. `jobs_unit_tests.go`...which I guess should be renamed `jobs_unit_test.go`.

Issue #1074, opened 2016-11-10
[Chains] make: refactor `func MakeAccounts()`


zramsay:
title is a comment from the linked PR #1072

zramsay:
not required for the Epic IMO but should come shortly thereafter

zramsay:
closed by #1235

PR #1235, opened 2017-02-19
Release 0.16 - restructure chains/maker to use definitions by eris-db

benjaminbollen:
After consolidating eris-cm and eris-pm into eris, there was a mixed use of definitions of eris-db types with respect to configuration file and genesis file.  This work re-aligns the two repositories and is dependent on https://github.com/eris-ltd/eris-db/pull/501; and should not be merged before.

checklist:
- [x] remove redundant definitions
- [x] restucture separation eris vs eris-db by introducing a `ErisDBAccountConstructor` contained in `chains/maker` which knowns `eris-db/genesis` types.
- [x] minor cleanup along the way
- [x] introduces `eris chains make --unsafe` to default to not writing `accounts.json` with the private keys to host
- [x] final review
- [x] update eris-db vendored version


chains/maker/accounts.go
```diff
@@ -1,97 +1,245 @@
...
 
 	"github.com/eris-ltd/eris/definitions"
 	"github.com/eris-ltd/eris/keys"
 	"github.com/eris-ltd/eris/log"
+
+	"github.com/eris-ltd/eris-db/genesis"
+	ptypes "github.com/eris-ltd/eris-db/permission/types"
 )
 
-func MakeAccounts(name, chainType string, accountTypes []*definitions.ErisDBAccountType) ([]*definitions.ErisDBAccount, error) {
```
zramsay: ++, closes #1074

chains/maker/accounts.go
```diff
@@ -1,97 +1,245 @@
...
+	// chain can be written to the host
+	// NOTE: [ben] because this is bad practice, it now requires explicit
+	// flag `eris chains make --unsafe` (unsafe bool in signatures below)
+	untypedPrivateKeyBytes []byte
+}
+
+// MakeAccounts specifies the chaintype and chain name and creates the constructors for generating
+// configuration, genesis and private validator files (the latter if required - for development purposes)
+// NOTE: [ben] if unsafe is set to true the private keys will be extracted from eris-keys and be written
+// into accounts.json. This will be deprecated in v0.17
```
zramsay: so we're introducing a flag that is planned to be deprecated? definitely not ideal & annoying for users

chains/maker/accounts.go
```diff
@@ -1,97 +1,245 @@
...
+			return nil, err
+		}
 
-		perms := &definitions.ErisDBAccountPermissions{}
-		var err error
-		if chainType == "mint" {
-			perms, err = ErisDBAccountPermissions(accountT.Perms, []string{}) // TODO: expose roles
-			if err != nil {
-				return nil, err
+		// NOTE: [ben] these auxiliary fields in the constructor are to be deprecated
```
zramsay: what if we hire another `[ben]`? initials is preferable IMO

chains/maker/accounts.go
benjaminbollen: I see your point; but rather my hope is that users never start using this flag.  If they do need the `accounts.json` then they need to add the flag for the mean time

chains/maker/accounts.go
benjaminbollen: I will veto the hiring of any other developer named Ben

definitions/do.go
```diff
@@ -27,6 +27,12 @@ type Do struct {
 	Dump           bool     `mapstructure:"," json:"," yaml:"," toml:","`
 	RemoteCompiler bool     `mapstructure:"," json:"," yaml:"," toml:","`
 	Save           bool     `mapstructure:"," json:"," yaml:"," toml:","`
+	// NOTE: [ben] the unsafe flag is introduced to start pushing out bad
+	// practices from the tooling with regards to extracting private keys
+	// from eris-keys.  Extracting the private keys can be convenient for
+	// the development and poc phase, but must be deprecated even in that
+	// case.
```
zramsay: i would remove this comment from this file and put it where I mentioned the comment should go in `cmd/chains.go`

definitions/do.go
```diff
@@ -112,9 +118,6 @@ type Do struct {
 	Service           *Service
 	ServiceDefinition *ServiceDefinition
 
-	// stuff from maker
-	Accounts []*ErisDBAccount
```
zramsay: :) :)

chains/maker/nodes.go
```diff
@@ -0,0 +1,75 @@
...
+			return fmt.Errorf("Unexpected behavior: constructor contains nil-GenesisAccount information.")
+		}
+	}
+
+	// generate the json bytes for the genesis file which are unique for all nodes.
+	genesisFileBytes, err := genesis.GenerateGenesisFileBytes(chainName, genesisAccounts, genesisValidators)
+	if err != nil {
+		return err
+	}
+	seedsString := strings.Join(seeds, ",") // format for config file (if len>1)
```
benjaminbollen: thanks for pointing out this issue!

definitions/do.go
benjaminbollen: yeah the `unsafe` bool moved around a few times and the comments got littered.

chains/maker/accounts.go
```diff
@@ -1,97 +1,245 @@
...
+// NOTE: [ben] if unsafe is set to true the private keys will be extracted from eris-keys and be written
+// into accounts.json. This will be deprecated in v0.17
+func MakeAccounts(name, chainType string, accountTypes []*definitions.ErisDBAccountType, unsafe bool) ([]*ErisDBAccountConstructor, error) {
+
+	accountConstructors := []*ErisDBAccountConstructor{}
+
+	switch chainType {
+	// NOTE: [ben] "mint" is a legacy differentiator that refers to the consensus engine that eris-db uses
+	// and currently Tendermint is the only consensus engine (chain) that is supported.  As such the variable
+	// "chainType" can be misleading.
```
VoR0220: WRT to this comment, perhaps changing the name of the signature would suffice here? "consensusEngine"?

benjaminbollen: I think from toolings perspective we want to keep it conceptually as a `chainType` and there is no harm in it being called "mint", just felt the comment was appropriate to clarify until we have more chain types.

PR #1242, opened 2017-02-23
Discussion: comment wrt accounts.json

benjaminbollen:
Should we have a mechanism to display UX changes once, but not every time?


zramsay:
I don't think we need one because we have less and less UX changes over time. It's not the end of the world if these warning linger a bit; we'll encounter them enough to remember to remove. FWIW the docs/tuts will be updated to include this anyways.

PR #1255, opened 2017-02-28
restructure so that we run through ALL of the tests

VoR0220:
Includes a sleeper function added to keep travis from stalling out, a placement of integration tests first in appveyor and travis runnings (so we can see if they actually run) followed by the unit testing. Also includes a minor formatting fixup for perform tests.

### DIFF
--- a/perform/perform_test.go
+++ b/perform/perform_test.go
@@ -1696,9 +1696,6 @@ func TestLogsSimple(t *testing.T) {
 		t.Fatalf("expected logs pulled, got %v", err)
 	}
 
-	fmt.Printf("MARMOT: %s\n", buf.String())
-	fmt.Printf("MARMOT BAD: %s\n", bufErr.String())
-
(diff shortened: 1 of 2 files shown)


perform/perform_test.go
```diff
@@ -1696,8 +1696,8 @@ func TestLogsSimple(t *testing.T) {
 		t.Fatalf("expected logs pulled, got %v", err)
 	}
 
-	fmt.Printf("MARMOT: %s", buf.String())
-	fmt.Printf("MARMOT BAD: %s", bufErr.String())
+	fmt.Printf("MARMOT: %s\n", buf.String())
+	fmt.Printf("MARMOT BAD: %s\n", bufErr.String())
```
zramsay: let's remove these, they were never meant to last

VoR0220: fair.

Issue #1262, opened 2017-03-07
[logs] unexpected behaviour & not well tested


benjaminbollen:
is part of #1230 

for release-0.16 `perform` fails:

```go test github.com/eris-ltd/eris/perform

Pulling default Docker images from quay.io
Pulling image 1 out of 3                image=quay.io/eris/data:0.16.0

Pulling image 2 out of 3                image=quay.io/eris/keys:0.16.0

Pulling image 3 out of 3                image=quay.io/eris/ipfs

Successfully pulled default images
Initializing default service definition files
Initialized Eris root directory          root=/tmp/eris




--- FAIL: TestLogsSimple (0.68s)
	perform_test.go:1699: expected certain log entries, got ""
Daemon is ready
ipfs  Yes /ipfs-f379f4a3-1d32-4fd4-b6ad-60b33556f2d0 0.0.0.0:4001->4001/tcp, 0.0.0.0:5001->5001/tcp, 0.0.0.0:8080->8080/tcp
Step 1 : FROM quay.io/eris/keys:0.16.0
 ---> 437dd5e31a11
Successfully built 437dd5e31a11
Step 1 : @^@%^@#%^&&#@% 
Step 1 : FROM ###^@%^@#%^&&#@%
FAIL
FAIL	github.com/eris-ltd/eris/perform	122.150s```

benjaminbollen:
perform/TestLogsSimple fails irrespective of the service used.

zramsay:
should address and test for #612

zramsay:
we're not going to be spending time fixing this unless user runs into a problem

PR #1263, opened 2017-03-07
Issue1230: silence perform logs simple

benjaminbollen:
this PR follows #1261 and makes `make test` go green by disabling one failing test; see #1262

### DIFF
--- a/perform/perform_test.go
+++ b/perform/perform_test.go
@@ -22,8 +22,8 @@ func TestMain(m *testing.M) {
 	// log.SetLevel(log.DebugLevel)
 
 	testutil.IfExit(testutil.Init(testutil.Pull{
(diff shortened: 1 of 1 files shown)


perform/perform_test.go
```diff
@@ -1656,10 +1656,12 @@ func TestPullBadName(t *testing.T) {
 	// }
 }
 
-func TestLogsSimple(t *testing.T) {
+// TODO: [ben] issue-1262: perform/TestLogsSimple fails
+// https://github.com/eris-ltd/eris/issues/1262
+func testLogsSimple(t *testing.T) {
```
zramsay: the usual way to silence a test is to prefix with an underscore ... lowercase works i suppose

PR #1267, opened 2017-03-15
Lookup service container IP address rather than relying on its hostname

### LINKED ISSUE #1228: Can't create a chain with 0.16.
Please include in your bug report:

- [ ] `docker version`

Docker version 1.13.1, build 092cba3

- [ ] `eris version`

Eris CLI Version: 0.16.0

- [ ] `go version` (if applicable)

go version go1.7.4 linux/amd64

- [ ] operating system details (osx/windows/linux)

Vagrant environment for release-0.16 (https://github.com/eris-ltd/eris-vagrant/commit/7acb03091f1b5db6f01b5dc56ecae1b989e68819)

- [ ] what you expected to happen

A chain to be made without error.

- [ ] what actually happened

```sh
Container interactive-db5c2429-0464-4bdc-870c-016060d03c06 exited with status 1
```

- [ ] steps to reproduction

```
eris chains make blockchain
```

silasdavis:
This should hopefully resolve an interaction with ISPs that DNS hijack and Alpine linux that does not seem to respect the lookup order between /etc/hosts and /etc/resolv.conf.

Regardless of that performing the lookup by inspecting docker is more robust than relying on chosen host name for the service to be resolvable. This mirrors how the chain container is found in 0.16.

Hopefully fixes #1228.

I've tested this locally by making a chain, starting it, and looking at the logs. Also `eris keys gen` works. Tools people please consider if this will have any unintended consequences because I'm not totally au fait with the eris codebase.

### DIFF
--- a/services/services.go
+++ b/services/services.go
@@ -140,8 +140,14 @@ func ExecService(do *definitions.Do) (buf *bytes.Buffer, err error) {
 			log.Info("exec_host not found in service definition file")
 			log.WithField("service", do.Name).Info("May not be able to communicate with the service")
 		} else {
+			// Grab the Container to inspect the service's IP address
+			cont, err := util.DockerClient.InspectContainer(main)
+			if err != nil {
+				return nil, util.DockerError(err)
+			}
 			service.Service.Environment = append(service.Service.Environment,
-				fmt.Sprintf("%s=%s", service.Service.ExecHost, do.Name))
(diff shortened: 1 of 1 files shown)


benjaminbollen:
looks good to me, but I will also wait for @zramsay to review whether this will interfere.

VoR0220:
This works. Merging.

PR #1280, opened 2017-03-28
add release checklist from /coding repo wiki, with commit history of whole wiki

zramsay:
- replaces #1225


benjaminbollen:
This should probably reflect that we work from release branches that sit between develop and master; but then overal structure is good!

PR #1283, opened 2017-03-31
Update vendor to match eris-db dependencies

silasdavis:
CLI currently has versions of eris-db, tendermint, and their dependencies that are not the same as those specified in the glide.lock. This does not for happy debugging make. It may also explain some serialisation compatibility issues.

I have updated the glide.lock to match the dependencies of current master (post-rename) eris-db and its dependencies.

I have also introduced a parallel `vendor.yml` file belonging to the govend tool (https://github.com/govend/govend). I have used govend in conjunction with this lock file to create a pruned vendor (see `cleanVendorGovend.sh`). My reasons for doing this are:

- I could not get glide-vc to work (it just fails with exit status 1 on my machine).
- Govend has built-in pruning of the vendor directory.
- Govend provides a more minimal 'lock the dependencies' feature than glide. Since we are only able to successfully use the locking facilities of glide (and various other Glide .operations are dangerous/destructive) it might provide less friction than to 'pretend' we are using glide as intended.
- Govend should allow you to update a single dependency without touching the entire tree and doing unwanted updates.
- Govend assume vendor.yml will be hand-maintained (though it can generate it from a scan and update individual dependencies) unlike glide.lock which glide likes to overwrite.

I have left the glide infrastructure in place, because I didn't want to bet on this too early, but if it proves to be a nicer solution we should probably switch to it exclusively (until Glide or Dep are mature enough to be usable as intended). It is easy enough to transform a glide.lock into a vendor.yml in a text editor (`:%s/  subpackages:\n\(  -.\+\n\)*//g` is a useful vim command for removing subpackage sections).


benjaminbollen:
cli has no direct dependency on tendermint anymore, does it?

silasdavis:
@benjaminbollen not direct, but transitive through a few deps chiefly eris-client. Here's the pruned dependency tree: 

https://gist.github.com/silasdavis/0fce4f40b7b54db6d3fa18f68140a836

VoR0220:
Closing in favor of #1286 . Yes there are more lines of coded added to the vendor. Yes I did a glide up. Yes we should keep these in there, because we want a up to date vendor. Yes the dependencies are cleaned out. Yes, the epm.yaml will be fixed, as soon as the vendoring is complete.

benjaminbollen:
What does the `epm.yaml` have to do with this, glide.yaml?

benjaminbollen:
I've tested this locally and works fine;  I will follow up with a PR to update eris-db to latest master to get needed hotfixes for release

PR #1284, opened 2017-04-01
make: add user story tests to `make test`

benjaminbollen:
make test runs:
- golang unit tests
- cm user story tests
- pkgs do user story tests

### DIFF
--- a/pkgs/abi/core.go
+++ b/pkgs/abi/core.go
@@ -8,11 +8,11 @@ import (
 	"strconv"
 	"strings"
 
+	ethAbi "github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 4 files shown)


pkgs/abi/core.go
```diff
@@ -8,11 +8,11 @@ import (
 	"strconv"
 	"strings"
 
+	ethAbi "github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/monax/cli/definitions"
 	"github.com/monax/cli/log"
 	"github.com/monax/cli/util"
-	ethAbi "github.com/ethereum/go-ethereum/accounts/abi"
```
VoR0220: Well this is definitely not right....should be fixed in #1286

PR #1287, opened 2017-04-03
add make test

benjaminbollen:
proposes same commit as in #1284; but moved off of my own release-0.16 branch to let it be independent.

`make test` runs:
1. check for gofmt formatting
2. build cli executable in /target folder
3. run golang unit tests
4. run chains make user story tests
5. run pkgs do user story tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,11 +55,11 @@ install_vendor:
 
 # build all targets in github.com/monax/cli
 .PHONY: build
-build:	check build_eris
+build:	check build_cli
 
 # build eris
 .PHONY: build_cli
-build_eris:
+build_cli:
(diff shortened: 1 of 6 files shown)


benjaminbollen:
I added a change to `./tests/test_jobs.sh` and `./tests/test_chains_make.sh` for the tests to use the latest build if available; if it does not exist, then it falls back to global `eris` executable, as was the behaviour before.

I have been caught too many times running the cli tests, only to find that I had not replaced the system wide executable for the purpose of testing.  This allows to run `make test` which will do `make build` which will be build the correctly named target executable for the test scripts to pick up in their tests.

It is still not perfect, but it is already a lot better to zero-effort local testing of cli.

PR #1295, opened 2017-04-04
Dockerfiles and lots of little fixes

compleatang:
Changes:

* use minor versions (rather than patch versions) of:
  * data
  * keys
* remove ipfs as a default image; change image to use their's
* s/eris/monax/g in a bunch of places (eris-keys needs work)
* make fmts (nice catching!)
* docker image build path collated (will needs jenkins work)
* base and build images now only produce minor version (which should be relied on across the stack)
* remove traces of remotes (for now)


chains/chains_test.go
```diff
@@ -201,7 +201,7 @@ func TestChainsNewConfig(t *testing.T) {
 }
 
 // chains start (--init-dir) should import the priv_validator.json (available in mint form)
-// into monax-keys (available in monax form) so it can be used by the rest
+// into eris-keys (available in monax form) so it can be used by the rest
```
compleatang: I reverted here to try to get chains maker tests passing, but they were failing on data imports and unfortunately I don't have time to debug why...

data/data.go
```diff
@@ -164,7 +164,7 @@ func ImportData(do *definitions.Do) error {
 
 		//required b/c `docker cp` (UploadToContainer) goes in as root
 		// and monax images have the `monax` user by default
-		if err := runData(containerName, []string{"chown", "--recursive", "monax", do.Destination}); err != nil {
+		if err := runData(containerName, []string{"chown", "-R", "monax", do.Destination}); err != nil {
```
compleatang: this is flaky sometimes on alpine

docker/x86/base/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM alpine:3.5
```
compleatang: Note this does not sha256 pin us like we were before. It also doesn't set it to alpine:edge. I think we're safer to stay with major release versions.

docker/x86/build/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM quay.io/monax/base:0.16
```
compleatang: note, the make file will need to control this

tests/test_chains_make.sh
```diff
@@ -146,7 +146,7 @@ run_test(){
     return 1
   fi
   dir_to_use=$chains_dir/$uuid/$direct
-  $cli_exec chains start $uuid --init-dir $uuid/$direct
+  $cli_exec chains start $uuid --init-dir $uuid/$direct --debug
```
compleatang: oops. please revert this before merging to release.

silasdavis:
This looks okay to me insofar as I'm able to test

PR #1300, opened 2017-04-06
Make sure monax.toml is written on fresh run of init

silasdavis:


initialize/initialize.go
```diff
@@ -56,6 +58,8 @@ func Initialize(do *definitions.Do) error {
 		return fmt.Errorf("Could not instantiate defaults:\n\n%s", err)
 	}
 
+	// TODO: [Silas] um, perhaps we should actually ask the filesystem what we've
+	// _really_ done rather than hard coding the output of tree...
```
zramsay: lol yeah good point.

PR #1301, opened 2017-04-06
pkgs: use ip, not hardcoded

### LINKED ISSUE #1281: cli assumes eris-keys will be found on http://172.17.0.2:4767
This will only work if keys was the first thing started, assumption introduced with: https://github.com/monax/cli/commit/f0e386ce8079a9c233bb27314a496e7331a24e2f

If no explicit `do.Signer` is provided then we need to do a docker inspect to find out where keys is.

zramsay:
- closes #1281


cmd/pkgs.go
```diff
@@ -88,6 +88,12 @@ func defaultSigner() string {
 	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
 		return "http://0.0.0.0:4767"
 	} else {
-		return "http://172.17.0.2:4767"
+		util.DockerConnect(false, "monax")
```
zramsay: this is not exactly ideal and we'll want to refactor the docker connect sequence. there are sufficient issues capturing some aspect of that refactor.

zramsay:
is green https://ci.monax.io/job/test_tools-cli-linux-x86-featurebranches/399/console even though GH hooks show red

PR #1314, opened 2017-04-07
Initialize int64s

VoR0220:
solves problem in tests where we missed some int64s


zramsay:
closes in https://github.com/monax/cli/commit/1f0fc08dc0f711fa3a9dddfea2b110054aa77a3a

PR #1317, opened 2017-04-08
core: remove 'files' command and ipfs integration

### LINKED ISSUE #1306: [monax files] remove command and IPFS integration
- adds unnecessary overhead and isn't used much if at all

zramsay:
- RIP ipfs; to be replaced with a  "working with ipfs" post using existing/maintained tools
- closes #1306


VoR0220:
Just want to add that this may come back in a different form, I'm going to need an "RPC suite" of some kind for package management purposes, of which IPFS definitely fits the bill for this usage. Might also be more useful in it's own docker container.

PR #1322, opened 2017-04-11
release fixes

zramsay:
- small re-name fixes for the release pipeline


zramsay:
todo drop choco support

misc/release/release.sh
```diff
@@ -86,15 +46,16 @@ LATEST_TAG=$(git tag | xargs -I@ git log --format=format:"%ai @%n" -1 @ | sort |
 MONAX_RELEASE=1
 
 # NOTE: Set these up before continuing:
-export GITHUB_TOKEN=
-export AWS_ACCESS_KEY=
-export AWS_SECRET_ACCESS_KEY=
+# export GITHUB_TOKEN=
```
silasdavis: Do we want these commented or not here at all?

If commented can we put a note to say why you would uncomment.

zramsay:
replaced by #1328 which needs the choco files removed

misc/release/release.sh
zramsay: yeah it's fine. either they'd be uncommented and filled out (but not ever committed), or whoever is running the script for release will have them set as env vars locally

compleatang: usage throws if they aren't filled out and tells the user to fill them in.

PR #1324, opened 2017-04-11
Update in line with website - not sure if you want to keep the eris commands for legacy

tombrushwood:
This might take some picking through, which I'm happy to do later in the week/next week


tombrushwood:
This PR is totally acting up... not showing changes already there. I'm closing it and I'll try again another time.

PR #1326, opened 2017-04-12
Update Chain maker Usage of Ints to Bool / Change monax namespace to hyperledger

VoR0220:
* This changes the usage of ints in golang to bools from the config files because that is what was being sent in anyway on the burrow side.
* This changes the namespace of monax/burrow to hyperledger/burrow. 

Note: relies on https://github.com/hyperledger/burrow/pull/569 and a subsequent vendor pull

### DIFF
--- a/initialize/default_files.go
+++ b/initialize/default_files.go
@@ -193,22 +193,22 @@ privileges should be given these keys.
 Usually this group will have the most number of keys of all of the groups.`
 		accountTypeDefinition.DefaultNumber = 25
 		accountTypeDefinition.DefaultTokens = 9999999999
-		accountTypeDefinition.DefaultBond = 0
-		accountTypeDefinition.Perms = map[string]int64{
-			"root":            0,
-			"send":            1,
-			"call":            1,
-			"create_contract": 0,
-			"create_account":  0,
-			"bond":            0,
-			"name":            1,
-			"has_base":        0,
-			"set_base":        0,
-			"unset_base":      0,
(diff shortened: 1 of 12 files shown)


initialize/default_files.go
```diff
@@ -319,21 +319,21 @@ similar to a network administrator in other data management situations.`
...
+			"create_account":  true,
+			"bond":            true,
+			"name":            true,
+			"has_base":        true,
+			"set_base":        true,
+			"unset_base":      true,
+			"set_global":      true,
+			"has_role":        true,
+			"add_role":        true,
+			"rm_role":         true,
```
zramsay: weren't these all changed to camelCase instead of snake_case?

zramsay: peruse the docs and update there too

VoR0220: we made a hack out of it so that we didn't have to do anything painful was my recollection.

VoR0220: the docs and functionality should stay the exact same so no need for doc updating idt

VoR0220:
@zramsay yup, they do, that's going to go in the vendor update. Just waiting on silas to merge that burrow bit.

VoR0220:
okay off to do that.

initialize/default_files.go
zramsay: yes, there is a need to update the doc https://github.com/monax/cli/blob/develop/docs/chain-making.md because otherwise a users will be confused

PR #1329, opened 2017-04-13
Provide clean vendor and remove GoVend

silasdavis:
In @VoR0220's updateVendorBurrow branch we have remnants of GoVend which I added as an experiment and because the cleanVendor.sh script (in particular glide-vc) wouldn't work for me. It seems to be working now. It is much better to remove this alternative vendoring than let it go out of sync with the glide.lock as it does on that branch. So this PR:

- Removes the GoVend script and vendor.yml
- Runs ./cleanVendor.sh (not sure why you hand't used this on the branch RJ, is it working for you?)


silasdavis:
Suggest this cleaned up version of vendor update is merged before (and into) https://github.com/monax/cli/pull/1327

zramsay:
works for me; compiles & starts a working chain

VoR0220:
We need to unmerge this. Im taking action. Glide is actively screwing our vendor at this point with transitive dependencies. Govendor works better. I just need to learn how to us it.

VoR0220:
As for glide vc...i have been running it...over and over and over again. Still largely polluted vendor.

Issue #1272, opened 2017-03-20
[chains] implement `chains make --output kubernetes`


compleatang:
**Problem** -- We want to be able to output k8s files as more users move to deploying into those clusters. Chains make has all of the information required to output appropriate configuration files that can then easily be consumed by a k8s cluster. 

**Solution** -- `chains make --output kubernetes` should output set of files that can be consumed by clusters. These should include the following:

* An `all_nodes.yaml` file that contains:
  * a `ConfigMap` with a condensed version of the genesis.json
  * a `Service` to the 1337 port on one of the nodes (probably commented by default) (random note, never load balance the 1337 port)
* A `keys.yaml` file that contains:
  * a range of `Secret`s which load the keys into Kubernetes secrets
* A `nodeXXX.yaml` file that contains:
  * a `Deployment` for the node which will start eris-db in "cluster mode" with the appropriate startup env vars
  * a `Service` which will expose the peer port within the cluster

zramsay:
partially implemented in burrow (https://github.com/hyperledger/burrow/tree/master/deployment) & this issue is labeled "discussion". If we do implement, the functionality should come from burrow first and the function imported/called from cli

zramsay:
aint gonna happen

PR #1331, opened 2017-04-13
deprecate the global [--machine] flag and cleanup Docker stuff

### LINKED ISSUE #1150: [Core] Remove `--machine` flag

zramsay:
- closes #1150, removing the `--machine` flag; to be replaced by some clustering feature using kubernetes (e.g., see #1272)
- connects to a machine named "default" if `eval` wasn't run putting the machine in scope (windows/osx)
- general cleanup of the `util/docker_client.go` file and the `DockerConnect` function; the latter especially should be carefully reviewed
- this PR should be tested locally on a mac


zramsay:
for background, see https://github.com/monax/cli/blob/master/docs/deprecated/using_docker_machine_with_eris.md

silasdavis:
Failing with `Monax requires [docker-machine] version >= 0.6.0`

cmd/monax.go
```diff
@@ -77,9 +77,7 @@ Complete documentation is available at https://monax.io/docs
 
 		// Compare `docker-machine` versions but don't fail if not installed.
 		dmVersion, err := util.DockerMachineVersion()
-		if err != nil {
-			log.Info("The marmots could not find [docker-machine] installed. While it is not required to be used with Monax, we strongly recommend it be installed for maximum blockchain awesomeness")
-		} else if !util.CompareVersions(dmVersion, dmVerMin) {
+		if !util.CompareVersions(dmVersion, dmVerMin) {
```
silasdavis: Does this mean that just because you have an older version of docker machine installed, we assume you want to use it?

What if I happen to have an older version installed but don't want to use it?

cmd/monax.go
```diff
@@ -77,9 +77,7 @@ Complete documentation is available at https://monax.io/docs
 
 		// Compare `docker-machine` versions but don't fail if not installed.
 		dmVersion, err := util.DockerMachineVersion()
```
silasdavis: Shouldn't we be checking the error value somewhere?

PR #1340, opened 2017-04-14
harmonize chains make output options

### LINKED ISSUE #1273: [Chains] Harmonize `chains make --output` options
**Problem**  -- `--zip` `--tar` are unnecessary cognitive overhead (also there is #1272). 

**Solution** -- These should all be combined into a string slice flag called `--output` which should contain the following nonexclusive options (once #1272 is shipped):

* `std` - display output on standard out (` ! --quiet`)
* `zip` - write zip files
* `tar` - write tarballs
* `kubernetes` - write k8s config files

zramsay:
- previous: `monax chains make --tar` or `monax chains make --zip` and the `--output` flag was default to true and basically required.
- now: `monax chains make --output zip` (or `tar` or, eventually, `kubernetes`)
- closes #1273, prepares for #1272
- the un-used `--force` flag for `monax chains start` was removed
- missed re-names


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,7 +1,7 @@
 Please include in your bug report:
 
 - [ ] `docker version`
-- [ ] `eris version`
+- [ ] `monac version`
```
silasdavis: typo: s/monac/monax

cmd/chains.go
```diff
@@ -282,7 +280,6 @@ func addChainsFlags() {
 	buildFlag(chainsStart, do, "ports", "chain")
 	buildFlag(chainsStart, do, "env", "chain")
 	buildFlag(chainsStart, do, "links", "chain")
-	chainsStart.PersistentFlags().BoolVarP(&do.Force, "force", "f", false, "force reinitialize the chain")
```
silasdavis: Are we removing this in favour of the more explicit `monax chains rm -f`?

docs/chain-making.md
```diff
@@ -1,4 +1,4 @@
----
+--i
```
silasdavis: typo: s/i/-

cmd/chains.go
zramsay: they were two different things. `chains start -f` would override an existing chain - `chains rm -f` simply removes it forcefully

silasdavis: yeah I meant if you want to start again the cycle of remove then start/re-init might be better. I realise `rm -f` is a different sense of force so I shouldn't have referenced that.

PR #1342, opened 2017-04-18
tests: remove old test scripts

zramsay:
- removes deprecated/outdated testing scripts
- removes `appveyor.yml` since we're dropping Windows support until Vagrant is supported
- testing is now handled exclusively by `make test` which does not touch any of the files deleted here


silasdavis:
How cleansing

Issue #1349, opened 2017-04-19
[services] remove some commands


zramsay:
toward the goal of cleaning up technical debt & reducing surface area, the following services sub-commands are being put on the chopping block; they do not provide essential functionality:
- `cat` (replace with `cat ~/.monax/services/keys.toml`
- `edit` (see path from cat)
- `update` (no added value)
- `ls` (handled by `monax ls`)
- `inspect` (use `docker inspect`)

zramsay:
closed by #1351

PR #1351, opened 2017-04-19
services: remove some sub-commands

zramsay:
- removes a handful of sub-commands from the `monax services` command
- deprecates: `cat | edit | ls | inspect | update`
- these sub-commands did not add value to the `monax` tool and/or their purpose can be accomplished via other means. see #1349 for more info.
- this begins a move toward deprecating the `monax service` command entirely in order to reduce the surface area of the `monax` tool

### DIFF
--- a/cmd/services.go
+++ b/cmd/services.go
@@ -6,8 +6,6 @@ import (
 	"strings"
 
 	"github.com/monax/cli/config"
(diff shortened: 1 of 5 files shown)


silasdavis:
Nice to see things go...

PR #1353, opened 2017-04-20
chains: remove some sub-commands

### LINKED ISSUE #1350: [chains] remove some commands
- related to #1349 
- on the chopping block:
- `inspect` => docker can do it (and better) ... (but keep `ip`)
- `restart` => just does a `stop/start`; redundant
- `ls` ... along with `services ls`, these commands can be handled by `monax ls`
- `cat`

zramsay:
- removes a handful for sub-commands for `monax chains`, similar to #1351 
- deprecates: `cat | inspect | ls | restart`
- these commands do not add value or can be accomplished otherwise. 
- closes #1350 and also severs the chain <=> service linking


cmd/chains.go
```diff
@@ -204,19 +177,6 @@ var chainsStop = &cobra.Command{
...
-var chainsInspect = &cobra.Command{
-	Use:   "inspect NAME [KEY]",
-	Short: "machine readable chain operation details",
-	Long: `display machine readable details about running containers
-
-Information available to the inspect command is provided by the
-Docker API. For more information about return values,
-see: https://github.com/fsouza/go-dockerclient/blob/master/container.go#L235`,
-	Example: `$ monax chains inspect simplechain -- will display the entire information about simplechain containers
-$ monax chains inspect 2gather Name -- will display the name in machine readable format
```
zramsay: lol, that was even before my time

compleatang: :doug:

tests/test_chains_make.sh
```diff
@@ -96,7 +96,7 @@ check_test(){
 
   # check genesis.json
   genOut=$(cat $dir_to_use/genesis.json | sed 's/[[:space:]]//g')
-  genIn=$($cli_exec chains plop $uuid genesis | sed 's/[[:space:]]//g')
+  genIn=$($cli_exec chains exec $uuid "cat /home/monax/.monax/chains/$uuid/genesis.json" | sed 's/[[:space:]]//g')
```
zramsay: this is why we can ax `cat/plop`, but need to keep `exec`. we can iterate towards using more docker and less cli-under-the-hood as well

PR #1359, opened 2017-04-21
chains: remove usage of mintkey

zramsay:
- removes `mintkey`, enabling deprecation of https://github.com/mint-client 
- goes with https://github.com/monax/keys/pull/104
- turns out mintkey was no longer even necessary


benjaminbollen:
LGTM

PR #1367, opened 2017-04-26
named jobs output files

### LINKED ISSUE #1033: [Pkgs] Support multiple output files for EPM pkgs command
Migrated from https://github.com/eris-ltd/eris-pm/issues/195

@j-h-scheufen commented on Wed Aug 17 2016

The change implemented in https://github.com/eris-ltd/eris-pm/issues/176 does not go far enough to support the needs of users who manage their epm jobs in multiple files. Currently EPM supports the `-f` flag which allows users to organize their jobs across multiple files; this is very useful and a requirement for bundles deployment, e.g. to be able to deploy contracts in separate phases, like libraries. However, it is currently not possible to control the output file and as far as I can see, the `jobs_output.json` would be overwritten on sequential jobs invocation.
## Requirements
1. Change the default behavior to use the input filename as a template for the output. `-f MyJobs.yaml` would result in `MyJobs.output.json`. Similarly `MyJobs.epm.yaml` -> `MyJobs.epm.output.json`. (exact logic to be specified)
2. Provide a commandline flag to control the output file name. There currently is a flag `-o`, but its description is unclear and it does not seem to be addressing this requirement.

zramsay:
- refactors some flags for `monax pkgs do`
- previously: `monax pkgs do --output` had two options: `json/csv`; this has changed to an optional string for naming the output file (used to be `jobs_output.json`) which will default now to `epm.output.json` (per #1033)
- previously the `monax pkgs do --file` flag could specify an `notepm.yaml` or whatever name and this did not affect the name of the output file; this has changed such that `monax pkgs do --file mypm.yaml` will, unless over-ridden with the `--output` flag, output a file called `mypm.output.json`
- the `monax pkgs do --dir` flag did not used to work: now it works
- the above functionality is captured in user-story tests that are considered seperate from the job fixtures
- closes #1033


cmd/pkgs.go
```diff
@@ -61,11 +59,6 @@ func addPackagesFlags() {
 
 func PackagesDo(cmd *cobra.Command, args []string) {
 	util.IfExit(ArgCheck(0, "eq", cmd, args))
-	if do.Path == "" {
-		var err error
-		do.Path, err = os.Getwd()
-		util.IfExit(err)
-	}
```
zramsay: moved to beginning of `pkgs.RunPackage` to accomodate `--dir` fix

pkgs/run_package.go
```diff
@@ -38,6 +67,14 @@ func RunPackage(do *definitions.Do) error {
...
 	}
 
+	if do.Path != gotwd {
+		for _, job := range do.Package.Jobs {
+			if job.Job.Deploy != nil {
+				job.Job.Deploy.Contract = filepath.Join(do.Path, job.Job.Deploy.Contract)
+			}
+		}
+	}
+
```
zramsay: this happens after loading the package; the contract path for a deploy job needs to be amended

tests/run_fixtures/simple/notpm.yaml
```diff
@@ -0,0 +1,84 @@
+jobs:
```
zramsay: duplicate of `epm.yaml`; used for testing ... `tests/run_fixtures/complex` will eventually be needed

tests/test_runner.sh
```diff
@@ -0,0 +1,239 @@
+#!/usr/bin/env bash
```
zramsay: this file is based on `test_chains_make.sh` and `test_jobs.sh`

cmd/pkgs.go
```diff
@@ -43,11 +40,12 @@ func addPackagesFlags() {
 	packagesDo.Flags().StringVarP(&do.ChainName, "chain", "c", "", "chain name to be used for deployment")
 	// TODO links keys
 	packagesDo.Flags().StringVarP(&do.Signer, "keys", "s", defaultSigner(), "IP:PORT of keys daemon which jobs should use")
-	packagesDo.Flags().StringVarP(&do.Path, "dir", "i", "", "root directory of app (will use $pwd by default)")              //what's this actually used for?
-	packagesDo.Flags().StringVarP(&do.DefaultOutput, "output", "o", "json", "output format which should be used [csv,json]") // [zr] this is not well tested!
-	packagesDo.Flags().StringVarP(&do.YAMLPath, "file", "f", "./epm.yaml", "path to package file which jobs should use")
+	packagesDo.Flags().StringVarP(&do.Path, "dir", "i", "", "root directory of app (will use $pwd by default)")
+	packagesDo.Flags().StringVarP(&do.DefaultOutput, "output", "o", "epm.output.json", "jobs output filename. will be named given the input filename, unless over-ridden")
```
VoR0220: Will be named...Actually this needs rewriting, it's kind of confusing the way it's currently written.

tests/test_runner.sh
VoR0220: This file I don't think needs to exist and should instead be handled by a unit test.

pkgs/jobs/writers.go
```diff
@@ -7,21 +7,13 @@ import (
...
-		return err
-	}
-
-	return os.Remove(setCsvPath())
-}
-
+// [zr] this should go (currently used by the nameReg writer)
 // WriteJobResultCSV takes two strings and writes those to the delineated log
 // file, which is currently epm.log in the same directory as the epm.yaml
 func WriteJobResultCSV(name, result string) error {
```
VoR0220: This will be taken out in the interface jobs and I will ensure that it does for name registry, it will merely be formed in the jobs_output.json

pkgs/jobs/job_manager.go
```diff
@@ -166,22 +166,27 @@ func defaultSetJobs(do *definitions.Do) {
...
-			}
-		}
-	case "json":
-		log.Info("Writing [jobs_output.json] to current directory")
-		results := make(map[string]string)
-		for _, job := range do.Package.Jobs {
-			results[job.JobName] = job.JobResult
-		}
-		return WriteJobResultJSON(results)
+	// check do.YAMLPath and do.DefaultOutput
```
VoR0220: This is going to be changed in the interfacedJobs...There's a better way to do this, and I intend to implement it that way. The name should instead be defaulting to a chain_id. We shouldn't allow users to tamper with this because then they could upset the output and this way we can give them more precise information about what chain these job outputs were deployed to.

That said...it works for now.

tests/run_fixtures/simple/notpm.yaml
VoR0220: these ought to be tested by unit tests, not necessarily integration tests/shell scripts I think, so it's a little overkill but either way, a test's a test, thanks for doing this.

PR #1370, opened 2017-04-27
Compilers rfc

VoR0220:
This contains the compilers RFC which dockerizes the ethereum solc compiler and utilizes it as a service and operates it mostly from the CLI.


initialize/default_files.go
```diff
@@ -15,6 +15,7 @@ const tomlHeader = `# This is a TOML config file.
 // ------------------ services ------------------
 
 var ServiceDefinitions = []string{
+	"solc"
```
zramsay: the missing `,` prevents the code as is from compiling.

initialize/initialize.go
```diff
@@ -140,13 +143,15 @@ func pullDefaultImages(images []string) error {
...
+		"solc":      version.ImageSolc,
 	}
 
 	for i, image := range images {
 		images[i] = versionedImageNames[image]
 
 		// Attach default registry prefix.
-		if !strings.HasPrefix(images[i], version.DefaultRegistry) {
+		// [rj] - this is very hacky, need to fix this either in images.go via a map, struct or a function
+		if !strings.HasPrefix(images[i], version.DefaultRegistry) && image != "solc" {
```
zramsay: it works though. `monax init` did exactly what it should've

version/images.go
```diff
@@ -20,4 +21,5 @@ var (
 	ImageKeys      = fmt.Sprintf("monax/keys:%s", KEYS_VERSION)
 	ImageDB        = fmt.Sprintf("monax/db:%s", DB_VERSION)
 	ImageCompilers = fmt.Sprintf("monax/compilers:%s", COMPILERS_VERSION)
```
zramsay: ummm, this needs to be deprecated.

zramsay:
the compilers tests also leave my git tree dirty

initialize/default_files.go
VoR0220: Shit. Sorry bout that.

initialize/initialize.go
VoR0220: It does. But something to clean in the future.

version/images.go
VoR0220: ++

VoR0220:
Ah shit...yea i will fix this.

VoR0220:
Monax/cli/compilers is a new package to maintain compilers. The monax/compilers needs to be rid from the vendor however. Good catch.

version/images.go
VoR0220: Well it will be deprecated. Thats for the second PR.

zramsay: what, why? `ethereum/solc` is *replacing* `quay.io/monax/compilers` ... that's definitely for one PR

VoR0220: There's a way ive created the compiler interaction with the deploy job that is for the second portion.

PR #1383, opened 2017-05-03
Version bump and vendor in Burrow develop dependencies

silasdavis:
This gets develop into a working state with respect to logging and ready for Burrow 0.17.0


silasdavis:
My god, I think I've fixed Jenkins

PR #1387, opened 2017-05-09
Add chain-url as an exposable flag

### LINKED ISSUE #1386: [pkgs] add --chain-url flag
When working against clusters we need to be able to give a raw RPC address for monax pkgs. This functionality previously existed in epm but was not carried over in the consolidation work.

compleatang:
Closes #1386

### DIFF
--- a/cmd/pkgs.go
+++ b/cmd/pkgs.go
@@ -38,6 +38,7 @@ in a package definition file`,
 
 func addPackagesFlags() {
 	packagesDo.Flags().StringVarP(&do.ChainName, "chain", "c", "", "chain name to be used for deployment")
+	packagesDo.Flags().StringVarP(&do.ChainURL, "chain-url", "", "", "chain-url to be used in tcp://IP:PORT format (only necessary for cluster and remote operations)")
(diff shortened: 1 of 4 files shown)


cmd/pkgs.go
```diff
@@ -38,6 +38,7 @@ in a package definition file`,
 
 func addPackagesFlags() {
 	packagesDo.Flags().StringVarP(&do.ChainName, "chain", "c", "", "chain name to be used for deployment")
+	packagesDo.Flags().StringVarP(&do.ChainURL, "chain-url", "", "", "chain-url to be used in tcp://IP:PORT format (only necessary for cluster and remote operations)")
```
VoR0220: This should default to a function grabbing the chain name, and from there we can create the nodeClient from it.

VoR0220:
Just need to change that default value, after that, this should be GTM.

cmd/pkgs.go
zramsay: that ends up being messy (happens with keysSigner() function or whatever it is.

PR #1389, opened 2017-05-11
chains make: add version flag for burrow's docker image to be used

### LINKED ISSUE #1378: [chains make] --version flag to write burrow/db version in config.toml

zramsay:
- requested by @j-h-scheufen
- closes #1378

### DIFF
--- a/chains/chains_test.go
+++ b/chains/chains_test.go
@@ -3,6 +3,7 @@ package chains
 import (
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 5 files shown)


silasdavis:
Now the place where this _may_ fall down in practice is when there is a breaking change of the Burrow RPC, logging, or other interface that crosses boundary into CLI through the Burrow vendoring into CLI.

Theory and practice suggest that this should happen infrequently relative to all changes in Burrow, so it's probably useful to be able to set the Burrow version when chains making. So I think I like this change, but it's not guaranteed to work. That said LGTM.

zramsay:
it's a fair point, however, we still control the default and that should never break. it will really only break by a user who attempts to be advanced by isn't

PR #1402, opened 2017-05-12
Compilers rfc

VoR0220:
- Deprecates monax compilers and utilizes the ethereum solc compiler via a docker container.


zramsay:
this will be implemented in many separate PRs

PR #1419, opened 2017-06-08
pkgs: fail if bad relation is used

### LINKED ISSUE #1135: [pkgs] abort jobs execution with error if invalid operators are used
Problem:
An `assert` job in EPM of the following example uses an invalid operator `neq` (correct: `ne`). The job logs as `Executing`, but there is no further output of the assertion and it proceeds with the next job.

Expectation: 
Using an invalid comparison operator should cause the job execution to fail with an error.

Jobs:
```
- name: note1EndorsedBefore
  job:
    query-contract:
      destination: $note1
      abi: PromissoryNote
      data: isEndorsed

- name: assertEndorsement1Before
  job:
    assert:
      key: $note1EndorsedBefore
      relation: neq
      val: "true"
```
Output:
```
Executing Job                                 note1EndorsedBefore
Return Value                                  true
Executing Job                                 assertEndorsement1Before
Executing Job                                 runEndorsement1
```

zramsay:
- adds a `default` to the `switch {}` for assert relations such that `monax pkgs do` will fail if an invalid relation is used. 
- implements a `expected-failure*` directory structure as part of the jobs_fixtures test sequence, so that we can test for, well, expected failures, such as the current fix. 
- closes #1135

### DIFF
--- a/pkgs/jobs/jobs_test_jobs.go
+++ b/pkgs/jobs/jobs_test_jobs.go
@@ -224,6 +224,8 @@ func AssertJob(assertion *definitions.Assert, do *definitions.Do) (string, error
 		} else {
 			return assertFail("<=", assertion.Key, assertion.Value)
 		}
+	default:
+		return "", fmt.Errorf("Error: Bad assert relation: \"%s\" is not a valid relation. See documentation for more information.", assertion.Relation)
 	}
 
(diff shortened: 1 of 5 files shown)


silasdavis:
Trying to retrigger PR build

PR #1420, opened 2017-06-09
deduplicate `job:` declaration from epm.yaml

### LINKED ISSUE #1061: [Pkgs] Get rid of `job` from epm.yaml in run files
This seems like something that doesn't need to be listed after every job name. We know it runs a job. So why repeat it after the name every time? Also could make our design more flexible as it relates to both #1034 and #1023 .

zramsay:
- closes #1061 
- **BREAKING CHANGE**:
- the `job:` field has been removed as extraneous and is no longer use in the `epm.yaml`
- to save compiled smart contract binaries, use `save: true`
- the `.abi` suffix should not be used


silasdavis:
LGTM, but wouldn't it make more sense to rename `Jobs` struct to `Job` as that seems more accurate, could  have `Jobs []*Job` instead of `Jobs []*Jobs` in:

```go
type Package struct {
	// from epm
	Account   string
	Jobs      []*Jobs
	Libraries map[string]string
}
```

PR #1421, opened 2017-06-09
pkgs: remove remote compiler functionality

### LINKED ISSUE #1346: [compilers] remove completely https://compilers.monax.io
From > 0.16.0 we will not be running a compilers service and so cli needs to reflect this.

zramsay:
- closes #1346 
- remote compiler no longer an option


silasdavis:
Cleansing...

PR #1433, opened 2017-07-13
Changed "AdvDeveloper" to "Advdeveloper" as the former never gets created.

### LINKED ISSUE #1428: [chains] AdvDeveloper not being made in tutorial

kyogesh91:
fixes #1428


docs/chain-making.md
```diff
@@ -164,7 +164,7 @@ Validator = 7
 
 ```
 
-You can see that we have zeroed out `Full` (which is a root + validator account type useful in simplechain scenarios) and `Developer` and utilized our new account type `AdvDevelop` which we will make three (3) of. The rest of the account types will use the defaults.
```
zramsay: nice catch on the typo

PR #1440, opened 2017-12-04
Update epm.yaml to be correctly parsed by CLI

guoger:
Job actions should be nested at the same level as job name,
instead of another level deeper under key `job`. Jobs defined
in such incorrect format are not parsed into valid jobs in
monax cli, therefore not executed.

- [x] `gofmt -s` your code
- [x] `git rebase develop` while on your feature/fix branch
- [x] use a descriptive title
- [ ] in the comments section, include the issue # (i.e., closes #577)
- [ ] provide a significant summary of the important changes the PR yields. Describe under-the-hood behaviour of code changes and/or modifications to the user experience

### DIFF
--- a/docs/contracts_deploying_adv/epm.yaml
+++ b/docs/contracts_deploying_adv/epm.yaml
@@ -1,49 +1,42 @@
 jobs:
 
 - name: deployGSFactory
-  job:
-    deploy:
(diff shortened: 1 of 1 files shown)


compleatang:
Thanks!

